Issue #20, opened 2019-07-29
Fix StreamTask/ArconTask template


Max-Meldrum:
implement some form of macro or trait to reduce code duplication and make it easier to reason about.

PR #21, opened 2019-07-30
[Rust] Adding: ArconEvent wrapper class for passing events and arcon_task macro

adamhass:
ArconEvent enum consisting of Element and Watermark for now, used for unified and simplified message passing within Arcon.

arcon_task macro implementing all the methods and traits which are identical on all task-components:
    Provide control_port, impl Actor, impl Require<PORT>, impl Send/Sync..
Task runners now only need to implement the methods handle_element and handle_watermark  and have the generic attributes IN (declaring the input type), and PORT (declaring the port type).

Resolving #20


arcc/arcon/src/streaming/channel/strategy/mod.rs
```diff
@@ -35,40 +29,33 @@ where
...
-            })?;
-
-            let timestamp = element.timestamp.unwrap_or(0);
-            if let Some(key) = key {
-                let keyed_msg = create_keyed_element(serialised_event, timestamp, key);
-                actor_path.tell(keyed_msg, source);
-            } else {
-                let element_msg = create_element(serialised_event, timestamp);
-                actor_path.tell(element_msg, source);
+        Channel::Remote(actor_path) => match event.to_remote() {
```
Max-Meldrum: I would suggest changing it to:
let msg = event.to_remote()?;
actor_path.tell(msg, source);

and let the caller handle the potential Error rather than printing it in the method itself.

arcc/arcon/src/streaming/task/flatmap.rs
```diff
@@ -1,45 +1,46 @@
...
 ///
-/// A: Input Event
-/// B: Port type for ChannelStrategy
+/// IN: Input Event
+/// PORT: Port type for ChannelStrategy
 /// C: Output Event
+#[arcon_task]
 #[derive(ComponentDefinition)]
-pub struct FlatMap<A, B, C>
+pub struct FlatMap<IN, PORT, C>
```
Max-Meldrum: Let's go with the output type as OUT rather than C for the tasks.

Max-Meldrum:
Looks good! :tada:

PR #43, opened 2019-09-03
Kafka Source/Sink added and Refactored Tasks

adamhass:
Two commits, don't squash them :) 

I've tried adding a ulimit increase to the ci yml, will see how that works with this PR. 

First Commit: Added Kafka integration. It's not at all optimized and it doesn't have well-defined behavior for committing reads/writes, but that will be added with State soon. The kafka integration requires the feature flag -kafka for it to be compiled/to run its tests. The tests are incomplete (no kafka mock to use) so no reason to add them to the CI pipeline yet, I'm thinking we add a kafka-integration test ci-pipeline when state is ready.

2nd Commit: Refactoring of our tasks so that we reduce code-duplication. I've called the new component "Node" but I'm not sure it's a good name? No need to rush to merging this PR. The point of this refactor is that the next steps of adding Epochs, separating input channels, buffers, etc. will be a lot easier and cleaner with this refactor since we decouple message handling from the tasks themselves.


Max-Meldrum:
Nice work! 😄 

Some cleanup would be good though, will comment on them.

execution-plane/arcon/src/streaming/source/kafka.rs
```diff
@@ -1,41 +1,281 @@
...
-    ctx: ComponentContext<KafkaSource<A>>,
-    target_pointer: ActorRef,
-    checkpoint_interval: u64,
-    _output_format: A,
+impl<OUT> Provide<ControlPort> for KafkaSource<OUT>
+where
+    OUT: 'static + ArconType + DeserializeOwned,
+{
+    fn handle(&mut self, event: ControlEvent) -> () {
+        match event {
```
Max-Meldrum: if let ControlEvent::Start = event {
   self.receive();
}

That should work. As the ControlEvent enum doesnt not have PartialEq, it doesn't work with a simple ==
or if

execution-plane/arcon/src/streaming/source/kafka.rs
```diff
@@ -1,41 +1,281 @@
...
+                &mut self.ctx
+            }
+            fn type_name() -> &'static str {
+                "EventTimeWindowAssigner"
+            }
+        }
+    }
+
+    #[test]
+    fn kafka_source() -> Result<()> {
```
Max-Meldrum: This test does not look like it is doing anything?
And it is sleeping 10 seconds

execution-plane/arcon/src/streaming/source/kafka.rs
```diff
@@ -1,41 +1,281 @@
...
+        attribute: i32,
+        location: Point,
+    }
+    #[arcon]
+    struct Point {
+        x: f32,
+        y: f32,
+    }
+    // JSON Example: {"id":1, "attribute":-13,"location":{"x":0.14124,"y":5882.231}}
+    mod sink {
```
Max-Meldrum: Is this needed? is it possible to use the Debug sink instead?

execution-plane/arcon/src/streaming/sink/kafka.rs
```diff
@@ -0,0 +1,169 @@
...
+        location: Point,
+    }
+    #[arcon]
+    struct Point {
+        x: f32,
+        y: f32,
+    }
+
+    #[test]
+    fn kafka_sink() -> Result<()> {
```
Max-Meldrum: Same here. Is this test doing anything?

execution-plane/arcon/src/streaming/sink/kafka.rs
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+}
+
+impl<IN> Provide<ControlPort> for KafkaSink<IN>
+where
+    IN: 'static + ArconType,
+{
+    fn handle(&mut self, event: ControlEvent) -> () {
+        match event {
```
Max-Meldrum: unnecessary match and error log.
if let ..  {
}

execution-plane/arcon/src/streaming/task/filter.rs
```diff
@@ -29,60 +19,89 @@ where
...
-        Ok(())
+impl<IN> Task<IN, IN> for Filter<IN>
+where
+    IN: 'static + ArconType,
+{
+    fn handle_watermark(&mut self, _w: Watermark) -> ArconResult<Vec<ArconEvent<IN>>> {
+        Ok(Vec::new())
+    }
+
+    fn handle_element(&mut self, element: ArconElement<IN>) -> ArconResult<Vec<ArconEvent<IN>>> {
```
Max-Meldrum: utilising the ArconResult

`let result = self.run_udf(&(element.data))?;`
`if result == 1 {
   Ok(vec!(ArconEvent::Element(element))
} else {
   Ok(Vec::new())
}
`

execution-plane/arcon/src/streaming/task/filter.rs
```diff
@@ -29,60 +19,89 @@ where
...
+        let mut result_vec = Vec::new();
+        let r1 = filter.handle_element(i1);
+        let r2 = filter.handle_element(i2);
+        result_vec.push(r1);
+        result_vec.push(r2);
+
+        let expected: Vec<i32> = vec![6];
+        let mut results = Vec::new();
+        for r in result_vec {
+            match r {
```
Max-Meldrum: using if let here to avoid unnecessary  empty ` {} ` cases

execution-plane/arcon/src/streaming/task/flatmap.rs
```diff
@@ -35,66 +27,91 @@ where
...
+where
+    IN: 'static + ArconType,
+    OUT: 'static + ArconType,
+{
+    fn handle_watermark(&mut self, _w: Watermark) -> ArconResult<Vec<ArconEvent<OUT>>> {
+        Ok(Vec::new())
+    }
+
+    fn handle_element(&mut self, element: ArconElement<IN>) -> ArconResult<Vec<ArconEvent<OUT>>> {
+        match self.run_udf(&(element.data)) {
```
Max-Meldrum: same here with `let result = ....?;`

execution-plane/arcon/src/streaming/task/flatmap.rs
```diff
@@ -35,66 +27,91 @@ where
...
+
+        let vec: Vec<i32> = vec![1, 2, 3, 4, 5];
+        let arcon_vec = ArconVec::new(vec);
+        let input = ArconElement::new(arcon_vec);
+
+        let exec_result = flatmap.handle_element(input);
+        let expected: Vec<i32> = vec![6, 7, 8, 9, 10];
+        let mut result_vec = Vec::new();
+
+        match exec_result {
```
Max-Meldrum: using if let here to avoid unnecessary  empty ` {} ` cases

execution-plane/arcon/src/streaming/task/map.rs
```diff
@@ -35,61 +26,89 @@ where
...
 
-    fn push_out(&mut self, event: ArconEvent<OUT>) -> ArconResult<()> {
-        let _ = self.out_channels.output(event, &self.ctx.system())?;
-        Ok(())
+impl<IN, OUT> Task<IN, OUT> for Map<IN, OUT>
+where
+    IN: 'static + ArconType,
+    OUT: 'static + ArconType,
+{
+    fn handle_element(&mut self, event: ArconElement<IN>) -> ArconResult<Vec<ArconEvent<OUT>>> {
```
Max-Meldrum: remove commented debug statement and use the `let result = ...?;` approach

execution-plane/arcon/src/streaming/task/map.rs
```diff
@@ -35,61 +26,89 @@ where
...
+        let r2 = map.handle_element(input_two);
+        let mut result_vec = Vec::new();
+        
+        result_vec.push(r1);
+        result_vec.push(r2);
+
+        let expected: Vec<i32> = vec![16,17];
+        let mut results = Vec::new();
+        for r in result_vec {
+            match r {
```
Max-Meldrum: using if let here to avoid unnecessary empty {} cases

execution-plane/arcon/src/streaming/task/node.rs
```diff
@@ -0,0 +1,111 @@
...
+                    }
+                } else {
+                    error!(self.ctx.log(), "Task failed to handle watermark");
+                }
+                self.output_event(ArconEvent::<OUT>::Watermark(*w)); // Forward the watermark
+            }
+        }
+    }
+    fn output_event(&mut self, event: ArconEvent<OUT>) -> () {
+        match self.out_channels.output(event, &self.ctx.system()) {
```
Max-Meldrum: unnecessary match statement.   doing a `if let Err(e) = .... {  error!() }` would be better

execution-plane/arcon/src/streaming/task/node.rs
```diff
@@ -0,0 +1,111 @@
...
+    }
+}
+
+impl<IN, OUT> Provide<ControlPort> for Node<IN, OUT>
+    where
+        IN: 'static + ArconType,
+        OUT: 'static + ArconType,
+{
+    fn handle(&mut self, event: ControlEvent) -> () {
+        match event {
```
Max-Meldrum: unnecessary match statement

execution-plane/arcon/src/streaming/source/kafka.rs
adamhass: Yeah I noticed this as well while working on the next commit. Already working on changing all the tests to use the DebugSink 👍

execution-plane/arcon/src/streaming/source/kafka.rs
adamhass: I was using the test to run "manual inspection tests" and figured it doesn't hurt to leave them as they were, useful for further development of them until we have a Kafka-mock or CI set-up with proper tests. It's not running when calling "cargo test" because it requires the feature flag to be added explicitly to the build/test command.

execution-plane/arcon/src/streaming/source/kafka.rs
adamhass: Is there any reason to not use the match? I think it's a better pattern for the ControlEvent because we may want to add features using other ControlEvents.

adamhass:
Added a 3rd commit. Adds the ArconMessage class, an envelope for storing sender information.
Updated codegen to match the new structure.
Restructured the protobuff messages.
A few fix-ups of the changes requested.

PR #50, opened 2019-10-16
Update to kompact 0.8.0

Max-Meldrum:


Bathtor:
Don't run any experiments on this with >32 threads (or use `KompactConfig::default()`). I forgot to export the necessary APIs that let you select a scheduler that works with more threads. Also there appears to be a bug in the scheduler for >64 threads (just noticed while running on AWS), so definitely don't do that :D Track https://github.com/Bathtor/rust-executors/issues/2 for updates on that issue. The API issue is fixed in https://github.com/kompics/kompact/commit/add1bb7f978a182cab071b69babd46d1b0f9dc9c but I don't want to publish a new version just for this, yet.

PR #53, opened 2019-11-26
Serialisation/Deserialisation additions

### LINKED ISSUE #49: Arcon Default Binary Format
With us moving away from the Weld runtime (#45), we will no longer be required to use Rust structs (although we could). We currently rely on Serde and Bincode to serialise our Rust structs into bytes so that they can be sent over the wire or stored in a state backend. However, I do not see this as a good approach moving forward. Why? Mainly two reasons: (1) lack of schema evolution; (2) performance reasons.

https://github.com/erickt/rust-serialization-benchmarks

**Note on Complexity** This is just based on my own experiences using the frameworks (In rust) on how complicated it would be to integrate them with Arcon.

| Framework       |  Schema Evolution  |  Performance |  Complexity |
| ------------- |-------------| -----| -----|
| Serde     | No | Medium  | Low
| Protobuf  | yes   |   Medium  | Medium
| Flatbuffers | yes   |    High  | High
| Cap’n Proto | yes   |   High | High

Both Flatbuffers and Cap'n Proto achieve similar performance due to not having any encode/decode step.

The point of this issue is to start a discussion.

Max-Meldrum:
This PR changes the following:

*  Deprecate arcon_messages
*  Adopt Protobuf as default storage format (motivated by #49)
*  Enable fast (unsafe) serialisation/deserialisation of in-flight data by using Abomonation (motivated by #49)
*  Add Unsafe and Reliable Kompact serialiser for **ArconMessage** 
* Add benches for serialisation/deserialisation
* Refactor + move remaining sinks to Operator impls

closes #49


Bathtor:
What do you mean by "heap allocated serialisation"?

Max-Meldrum:
What I mean is that you currently have to provide a correct size_hint for the "Serialiser". Which is not possible if you have something like a Vec<_> in your struct. The BytesMut is just gonna do a with_capacity(size_hint) and may therefore panic.

Max-Meldrum:
Well, I realise now that I can redesign it. brb. :beer:

PR #60, opened 2020-01-15
Fixing issues pointed out by clippy

mrobakowski:
Also removed a few `unsafe impl`s


mrobakowski:
Sure! 

> cargo --benches --all is failing on the CI

I forgot to test if it compiles with all the features, I'm gonna fix that tomorrow.

mrobakowski:
Now it compiles on my machine. I even went one step further and did `cargo check --all --benches --all-features`, and that also compiles, so it should be all good now.

PR #77, opened 2020-02-26
benches and additional checks for allocator

Max-Meldrum:


execution-plane/arcon/src/allocator/mod.rs
```diff
@@ -45,17 +44,25 @@ impl ArconAllocator {
...
         let layout = Layout::from_size_align_unchecked(required_bytes, align);
         let mem = System.alloc(layout);
+
+        if mem.is_null() {
+            return arcon_err!("{}", "SystemAllocator is OOM");
+        }
+
         self.curr_alloc += layout.size();
+        let id = self.alloc_counter;
         self.alloc_counter += 1;
```
Bathtor: This could theoretically overflow in a very long running streaming pipeline with many allocations and deallocations. Should be more specific as to the failure behaviour?

Max-Meldrum: How about AllocId = (u64,u64) where (epoch, id) and id reaches u64::MAX_SIZE, then we simply bump the alloc epoch and start from 0 in id again?

Bathtor: Works. We could also try to keep track of the lowest active allocation and wrap around if it's ok to do so and error otherwise. It really depends on how many active allocations we want to support and how likely we think it is that they are long lived.
I don't really have a particular opinion on the matter. I just wanted to be sure that we are not silently ignoring an overflow, since in release Rust they don't generate a panic.

mrobakowski: `u128` is also a thing in Rust :)

Bathtor: That, too^^

PR #78, opened 2020-02-28
EventBuffer & BufferPool

### LINKED ISSUE #74: Reusable message buffers for ChannelStrategies
Preallocate message buffers through #57 at start and send message by message if there is no available buffer in the ChannelStrategy and that we are not able to allocate more through the allocator.

Max-Meldrum:
EventBuffer backed by the ArconAllocator and a BufferPool of reusable EventBuffers...

Not integrated into the streaming logic yet. But uploading early PR for feedback etc... 

Will close #74 once integrated...


execution-plane/arcon/src/data/buffer.rs
```diff
@@ -0,0 +1,295 @@
...
+use arcon_error::*;
+use crossbeam_utils::CachePadded;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc, Mutex};
+
+/// A reusable buffer allocated through [ArconAllocator]
+///
+/// Assumes a Single-writer, single-reader setup.
+#[derive(Debug)]
+pub struct EventBuffer<T> {
```
mrobakowski: The `Drop` impls of `T`s contained within are never called, right? Might want to document that.

mrobakowski: Potentially related: http://cglab.ca/~abeinges/blah/everyone-poops/

Max-Meldrum: This would only be a problem if the EventBuffer is dropped before the BufferReader is consumed. As the BufferReader.into_iter() call will consume all T's and take ownership and thus drop them once they go out of scope.

Max-Meldrum:
Should probably try to get this merged at some point. Only tricky part left is integrating it nicely with ArconMessage and its requirements (e.g., Protobuf, Abomonation).

```rust
// current
pub struct ArconMessage<A: ArconType> {
    /// Buffer of ArconEvents
    #[prost(message, repeated, tag = "1")]
    pub events: Vec<ArconEventWrapper<A>>,
    /// ID identifying where the message is sent from
    #[prost(message, required, tag = "5")]
    pub sender: NodeID,
}
// new?
pub struct ArconMessage<A: ArconType> {
    /// Buffer of ArconEvents
    #[prost(message,  tag = "1")]
    pub events: BufferReader<ArconEventWrapper<A>>,
    /// ID identifying where the message is sent from
    #[prost(message, required, tag = "5")]
    pub sender: NodeID,
}
```

Max-Meldrum:
Some changes since last commit:

1. All channel strategies now use a BufferPool of buffers allocated through the ArconAllocator. Their implementations at this point are very basic (i.e. busy waits for a new buffer).

2. Fixed the ChannelBroken error ouput

3. There is now an ArconMessage and a RawArconMessage, where the latter supports serialisation and can be sent over the wire or stored in a state backend when having to temporarily block channels etc... 

```rust
#[derive(Debug, Clone)]
pub struct ArconMessage<A: ArconType> {
    /// Batch of ArconEvents backed by an [EventBuffer]
    pub events: BufferReader<ArconEventWrapper<A>>,
    /// ID identifying where the message is sent from
    pub sender: NodeID,
}

#[cfg_attr(feature = "arcon_serde", derive(Serialize, Deserialize))]
#[derive(PMessage, Clone, Abomonation)]
#[cfg_attr(feature = "arcon_serde", serde(bound = "A: ArconType"))]
pub struct RawArconMessage<A: ArconType> {
    /// Batch of ArconEvents
    #[prost(message, repeated, tag = "1")]
    pub events: Vec<ArconEventWrapper<A>>,
    /// ID identifying where the message is sent from
    #[prost(message, required, tag = "2")]
    pub sender: NodeID,
}
```
Both structs may be processed through the common function:

```rust
/// Iterate over a batch of ArconEvent's
#[inline]
fn handle_events<I>(&mut self, sender: NodeID, events: I) -> ArconResult<()>
where
     I: IntoIterator<Item = ArconEventWrapper<OP::IN>>,
{
      'event_loop: for event in events.into_iter() {
      }
}
```

5. Moved recovery_tests into src/test as it is more of an internal test, also that it depends on functions that are now under a cfg(test) flag.

6. Deleted node.rs under benches/.  couldn't see that it "tests" anything specific. It acted very similar to existing Kompact experiments. For Arcon, we can instead focus on NEXMark experiments.

7. Modified test cases to now use ArconPipeline rather than directly building a KompactSystem.

8. Added a buffer module under src/.  Currently it just has ``event``, but later on, network (kompact) buffers backed by the ArconAllocator should be put there as well.

Max-Meldrum:
I think it is ready to be merged. Improvements to serialisation and channel strategies can be fixed in new PRs.

PR #79, opened 2020-02-28
State backend infrastructure

mrobakowski:
# State Backends
This PR introduces a few important traits and types related to state management. Every concrete state backend type must adhere to the `StateBackend` trait. https://github.com/cda-group/arcon/blob/b1e5cebee5241b6d88a64e10f88952e7b951e67e/execution-plane/arcon/src/state_backend/mod.rs#L8-L23 This trait is pretty limited and of little use by itself. 

# State Handles
Next, we have several different state handle types. Heavily inspired by Flink. The base trait is simply named `State<SB, IK, N>`. https://github.com/cda-group/arcon/blob/b1e5cebee5241b6d88a64e10f88952e7b951e67e/execution-plane/arcon/src/state_backend/state_types/mod.rs#L42-L52 `SB` is the concrete type of a state backend, e.g. `RocksDb` or `InMemory`. State handle by itself is unable to access the state in any way, so every operation takes a (possibly mutable) reference to the `SB` that the handle originated from.

`IK` and `N` are, respectively, types of the current item key and the current namespace. The idea is that the state handles transparently keep track of those, so the per-operator business logic can transparently handle grouping by key and windowing. The namespace is usually just a window index and maybe it should be just renamed to that. I was also thinking about whether we should keep the tracking of the current item key and namespace as a part of the state handle type, or should we pass them explicitly to every operation.

Concrete implementations of the state handles for every state backend are usually also parametrized by the serializers used for serialization and deserialization of keys and values.
https://github.com/cda-group/arcon/blob/b1e5cebee5241b6d88a64e10f88952e7b951e67e/execution-plane/arcon/src/state_backend/rocks/map_state.rs#L34-L42 `KS` stands for key serializer and `TS` for value serializer. Those are often not the same, because sometimes we require an extra property from the key serializer - that it serializes a certain type to byte arrays of certain fixed length. This was motivated by making it possible to utilize the bloom filters built-in in RocksDB for quickly figuring out which values belong to which item key/namespace by looking up a known prefix. https://github.com/cda-group/arcon/blob/b1e5cebee5241b6d88a64e10f88952e7b951e67e/execution-plane/arcon/src/state_backend/rocks/mod.rs#L589-L602

# State Builders
As you saw earlier `StateBackend` doesn't have any methods to create a `State` state handle. This is where state builders come in. Here's an example for `ValueState`s: https://github.com/cda-group/arcon/blob/b1e5cebee5241b6d88a64e10f88952e7b951e67e/execution-plane/arcon/src/state_backend/builders.rs#L69-L79 Those traits are pretty much always implemented for the types that `StateBackend` is implemented for, but the traits itself are not part of the `StateBackend` trait, because they are not object-safe (it will become clear later why this is important). Here you can see usage example of a state builder: https://github.com/cda-group/arcon/blob/b1e5cebee5241b6d88a64e10f88952e7b951e67e/execution-plane/arcon/src/state_backend/mod.rs#L143-L172 The trait bounds become pretty complicated very quickly. That's why I've tried to make `dyn StateBackend`s also very usable. All the builder traits are implemented for `dyn StateBackend`, given that the parameters satisfy the union of the trait bounds of all the builder impls for concrete `StateBackend`s. This is done by the slightly scary-looking macro at the top of the `builders.rs` file.

If you look closely enough, you'll notice that the state handle type produced by the dynamic builders is in the form of `Box<dyn State<dyn StateBackend, ...>>` - usually the `SB` parameter is a concrete `StateBackend`, here it is dynamic. There's a special wrapper type `WithDynamicBackend<SB, C>` that downcasts a `dyn StateBackend` to the type that the underlying `State` expects. https://github.com/cda-group/arcon/blob/b1e5cebee5241b6d88a64e10f88952e7b951e67e/execution-plane/arcon/src/state_backend/state_types/boilerplate_impls.rs#L11-L15 The file above also contains impls for some boxed versions of state handles, because in Rust `Box<dyn Trait> : Trait` is not satisfied by default. With that pile of boilerplate, the usage of dynamic state backends becomes quite pleasant. https://github.com/cda-group/arcon/blob/b1e5cebee5241b6d88a64e10f88952e7b951e67e/execution-plane/arcon/src/state_backend/mod.rs#L77-L94 Notice the lack of _any_ trait bounds in the function signature above. There's an even nicer API for building states à la "builder API" pattern that assumes some defaults. It's implemented at the bottom of `builders.rs` file.

# Serialization
The file `serialization.rs` file abstracts over `prost` and serde-based `bincode`. It also implements a bare-bones serialization method that just dumps bytes of several primitive types with specified endianness. `Prost` is ok for serializing values in most cases, but fails the requirement of fixed-size serialization of simple types, so it cannot be used for key types in state handles. 

By default the builder API assumes `Prost` as the value serializer and `LittleEndianBytesDump` as the key serializer. If `arcon` is built with the `arcon_serde` feature flag, both defaults become `Bincode`. Personally I find serde and bincode much nicer to work with than prost, but I guess that ship has already sailed.

# Implemented State Backends
There are currently two implemented state backends - `InMemory` and `RocksDb`.

The `InMemory` state backend is equivalent to `HashMap<Vec<u8>, Box<dyn Any>>` and never serializes the actual values. This state backend cannot be checkpointed and "restoring" it always produces an empty state backend.

The `RocksDb` state is based on rocksdb. Checkpointing and restoring (to/from a local directory) works. You can see an end-to-end example in [`tests/recovery_test.rs`](https://github.com/cda-group/arcon/blob/b1e5cebee5241b6d88a64e10f88952e7b951e67e/execution-plane/arcon/tests/recovery_tests.rs).


mrobakowski:
Uh, this was supposed to be a draft PR, because I still have a few details to fix and clean up, but apparently prepending the PR name with `WIP:` doesn't work on github as it does on gitlab ¯\_(ツ)_/¯

Max-Meldrum:
Nice work! :) As you said the PR is not finished, I'll wait a bit more until I go through stuff in more detail. But I may leave comments on general stuff.

execution-plane/arcon/Cargo.toml
```diff
@@ -24,13 +24,18 @@ arcon_macros = { version = "0.1", path = "arcon_macros" }
...
+static_assertions = "1.1.0"
+itertools = "0.8.2"
+smallbox = { version = "0.8", features = ["coerce"] }
 
 # Serialisation
 abomonation = "0.7.3"
 abomonation_derive = "0.5.0"
 prost = { git = "https://github.com/Max-Meldrum/prost.git", branch = "derive_generics" }
 prost-types = "0.6"
+serde = { version = "1.0.104", features = ["derive"] }
```
Max-Meldrum: Would it be possible to keep serde/bincode as a feature? Or is something very dependent on serde? e.g., serialisation of timers?

execution-plane/arcon/src/lib.rs
```diff
@@ -8,6 +8,7 @@
 //! [arcon_codegen] to generate the Rust code.
 
 #![feature(unboxed_closures)]
+#![feature(fn_traits)]
```
Max-Meldrum: I'm not up to date with these nightly features. So my questions: (1) Are they necessary? (2) What are their status? Will they most likely end up in stable in the coming future?

Now with Kompact on stable, it would be nice to aim for stable on arcon as well :smile:

execution-plane/arcon/Cargo.toml
```diff
@@ -24,13 +24,18 @@ arcon_macros = { version = "0.1", path = "arcon_macros" }
 arcon_error = { version = "0.1", path = "arcon_error" }
 arcon_extra = { version = "0.1", path = "arcon_extra" }
 kompact = { git = "https://github.com/kompics/kompact.git", branch = "master"}
-uuid = "0.8"
+uuid = { version = "0.8", features = ["serde", "v4"] }
+static_assertions = "1.1.0"
+itertools = "0.8.2"
+smallbox = { version = "0.8", features = ["coerce"] }
```
Max-Meldrum: Did not know about smallbox. nice :)

execution-plane/arcon/src/lib.rs
Bathtor: I mean...I'll continue focusing on nightly Kompact.
Just because Kompact works on stable, doesn't mean I like it that way^^

Max-Meldrum: boooo

Bathtor: I'm not really sure that the point of the `Fn*` traits is, though...all it gives you a syntax like Scala's `apply` and the ability to pass custom types in places that expect closures without double wrapping (which the compiler should eliminate anyway). They seem very hacky...not sure they'll ever hit stable due to combination of hacky vs. utility...

execution-plane/arcon/Cargo.toml
mrobakowski: I think I may be able to make everything use Prost, so serde would stay optional. That's one of the reasons I still have the PR as WIP 😄

execution-plane/arcon/src/lib.rs
mrobakowski: I _thought_ we needed it to use the nice fn-trait-like syntax for our `SafelySendableFn`. Turns out I was mistaken.

mrobakowski: We only need `unboxed_closures` for that, not `fn_traits`.

mrobakowski:
I've updated the PR message. The failed check doesn't happen on my machine and tbh I have no idea why it's happening.

execution-plane/arcon/src/lib.rs
```diff
@@ -34,12 +29,11 @@ pub mod util;
 #[cfg(test)]
 pub mod test_utils {
     use crate::allocator::ArconAllocator;
+    use once_cell::sync::Lazy;
```
Max-Meldrum: Any special reason for the swap between lazy_static and once_cell?

mrobakowski: It's nice IMO. https://docs.rs/once_cell/1.3.1/once_cell/#faq

execution-plane/arcon/src/util/prost_helpers.rs
```diff
@@ -0,0 +1,21 @@
+use prost::Message;
+
+// it's so stupid that this is needed
```
Max-Meldrum: This comment doesn't really add anything ^^
A short comment on why it is needed would be better.

execution-plane/arcon/src/stream/operator/mod.rs
```diff
@@ -15,17 +18,56 @@ where
...
+    fn handle_epoch(
+        &mut self,
+        epoch: Epoch,
+        ctx: OperatorContext<OUT>,
+    ) -> Option<ArconResult<Vec<u8>>>;
+}
+
+pub struct OperatorContext<'c, 's, OUT: ArconType> {
+    channel_strategy: &'c mut ChannelStrategy<OUT>,
+    pub state_backend: &'s mut dyn StateBackend,
```
Max-Meldrum: Wondering how we'll deal with stateless operators/nodes. They won't need any StateBackend. If this should be an option or something.

execution-plane/arcon/src/util/prost_helpers.rs
mrobakowski: Whoops, sorry. I was using `Option<T>` as a state in some test, and prost doesn't really do well with most std types.

execution-plane/arcon/src/stream/operator/mod.rs
Bathtor: Doesn't necessarily hurt to give them one anyway, even if they are ignoring it, does it?

mrobakowski: `Node` always has its own `StateBackend` though. It can pass it without any problems to the operator inside, even though the operator doesn't use it.

Max-Meldrum: Well, I guess not ^^

execution-plane/arcon/src/util/prost_helpers.rs
mrobakowski: Ah, nevermind. It wasn't even a test. It was needed to port `IncrementalWindow` to use state backends.

execution-plane/arcon/Cargo.toml
```diff
@@ -25,34 +26,39 @@ arcon_error = { version = "0.1", path = "arcon_error" }
...
 rdkafka = { version = "0.23", optional = true }
 tokio = { version = "0.2", optional = true, features = ["full"] } # TODO: figure out which are truly needed
 tokio-util = { version = "0.2", optional = true, features = ["full"] }
-bytes = { version = "0.5", optional = true }
 futures = { version = "0.3", optional = true }
 serde_json = { version = "1.0.44", optional = true }
 serde = { version = "1.0.104", features = ["derive"], optional = true }
-rocksdb = { git = "https://github.com/tikv/rust-rocksdb.git", optional = true }
+bincode = { version = "1.2.1", optional = true }
+rocksdb = { git = "https://github.com/mrobakowski/rust-rocksdb.git", branch = "arcon", optional = true }
```
Max-Meldrum: I saw you mentioned that there are breaking changes in the PR on the rust-rocksdb repo. And the tests are not passing. Just concerned if this change will be integrated or not. I mean it is kinda the same with the prost branch. But the less custom branches the better :D 

If the changes are not merged, would it easy to adjust the current arcon code to work with the rust-rocksdb master?

mrobakowski: tikv version doesn't support checkpoints and here's my pull request for the one on crates.io https://github.com/rust-rocksdb/rust-rocksdb/pull/385. I _think_ it would be possible to make it work with the current version of rust-rocksdb, but it would have to use mutable global state and I don't really want to touch stuff like that.

execution-plane/arcon/tests/recovery_tests.rs
```diff
@@ -0,0 +1,238 @@
...
+                &runtime_dir,
+            )
+            .unwrap(),
+        ),
+        None => Box::new(RocksDb::new(&runtime_dir).unwrap()),
+    }
+}
+
+/// CollectionSource -> Window -> Map -> LocalFileSink
+fn run_pipeline(sink_path: &str) -> Result<(), ()> {
```
Max-Meldrum: Nice :)

execution-plane/arcon/tests/recovery_tests.rs
```diff
@@ -0,0 +1,238 @@
...
+    if window_node.is_faulty() || map_node.is_faulty() || file_sink_node.is_faulty() {
+        return Err(());
+    }
+
+    Ok(())
+}
+
+#[test]
+fn run_test() {
+    let test_directory = "./recovery_test";
```
Max-Meldrum: If the directory is not needed after, then we should use tempfile::tempdir. Or if there is a reason for it to remain after the tests, then add it to gitignore.

Bathtor: I recommend something like [Fixture](https://github.com/kompics/kompact/blob/a9b1a5b9909343e4992ade96961cfc33b2353c4a/core/src/lib.rs#L289-L337) actually...seems to work rather well.

mrobakowski: Oh yeah, no reason. When I first wrote it I just wanted to inspect the folder by hand for some reason and forgot to switch to `tempfile`.

Max-Meldrum:
Otherwise, is there anything you feel like is left to add? Or shall we move towards merging the PR.

I think it is adds a solid foundation which we can add more on top of later.

mrobakowski:
I think it's ready to merge.

Max-Meldrum:
Had another look and I am wondering what the SerializbleFixedSize and the byte_dump macros are used for (Native, BigEndian, LittleEndian)?

mrobakowski:
`SerializableFixedSizeWith<S>` trait is for types that guarantee that they will always serialize into exactly `SIZE` bytes using serializer `S`. This allows us to use fixed-size prefixes in rocks to populate the bloom filters. `Bincode` serializer guarantees that for a lot of primitives, but `Prost` doesn't do that at all. Even integers are encoded in a variable-length fashion. So, since `Bincode` is behind an optional feature, I implemented a simple, byte-dumping serializer that can be used with primitives, `Vec<T>`, `&[T]`, `String` and `&str`.

PR #84, opened 2020-04-07
Add sled- and FASTER-based state backend implementations

mrobakowski:
This PR adds state backend implementations based on [sled](https://github.com/spacejam/sled) and [FASTER](https://github.com/microsoft/FASTER). The recovery tests are also implemented for them. 

Sled checkpointing and recovery are implemented through full export/import for now, because sled doesn't support efficient snapshots/MVCC yet (there are plans for it afaik). 

FASTER doesn't support range scans, so map state is implemented with an additional storage entry that contains an array of all the keys related to the particular map state instance.


Max-Meldrum:
Looks good! Is it ready to be merged?

mrobakowski:
I think so :). I'm just waiting for the CI to finish.

mrobakowski:
`test_faster_recovery_pipeline` did not pass, again. And I don't really know why, since some logs probably got swallowed :/ and I cannot reproduce this on my machine.

mrobakowski:
Still no idea why it doesn't pass. We could just ignore this test for now...

Max-Meldrum:
Sure, just comment it out or something.

PR #86, opened 2020-04-08
Reschedule CollectionSource occasionally 

Bathtor:
Reschedule CollectionSource occasionally (every 500k elements) so it doesn't execute a single handler forever. 

Kompact doesn't deal well with handlers hanging on to their threads forever. In particular not for the `ControlEvent` handler, since there are internal things that also need to happen in response to these.

### DIFF
--- a/execution-plane/arcon/src/stream/source/collection.rs
+++ b/execution-plane/arcon/src/stream/source/collection.rs
@@ -5,15 +5,28 @@ use super::SourceContext;
 use crate::data::ArconType;
 use kompact::prelude::*;
 
-#[derive(ComponentDefinition)]
+const RESCHEDULE_EVERY: usize = 500000;
+
+#[derive(Debug, Clone, Copy)]
+struct ContinueSending;
+struct LoopbackPort;
+impl Port for LoopbackPort {
+    type Indication = Never;
+    type Request = ContinueSending;
+}
+
+#[derive(ComponentDefinition, Actor)]
 pub struct CollectionSource<IN, OUT>
 where
     IN: ArconType,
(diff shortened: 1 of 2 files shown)


Max-Meldrum:
Looks good! Might as well remove the commented code though :)

execution-plane/arcon/src/stream/source/collection.rs
```diff
@@ -52,21 +72,35 @@ where
...
+where
+    IN: ArconType,
+    OUT: ArconType,
+{
+    fn handle(&mut self, _event: Never) {
+        unreachable!("Never type has no instance");
+    }
+}
+
+//ignore_indications!(LoopbackPort, CollectionSource);
```
Bathtor, line 102: Any idea why that doesn't work on Arcon? You pointing to an old Kompact version where it wasn't available, yet?

Max-Meldrum: Was it added recently? Could be that we need to do a cargo update (as we are currently using git master as dependency)

Bathtor: I mean...a few weeks ago...I don't know when you did cargo update last.

Anyway, if you do it at some point, try of that works, maybe and if yes delete the other variant. I can't remember if I once again forgot about supporting components with those pesky generic type parameters :D

Bathtor:
Oehm...that didn't fail before. Flaky test?

Max-Meldrum:
> Oehm...that didn't fail before. Flaky test?

flaky. Will add an issue to fix socket related tests

PR #96, opened 2020-04-21
State backend metrics

mrobakowski:
This PR adds a wrapper `StateBackend` called `Metered` that tracks metrics of state backend operations of any other state backend.


Max-Meldrum:
Looks nice :) I'm not fully sure I understand how it is supposed to be used though. Is it just meant for unit tests and checking the performance differences?

mrobakowski:
Yes, this is going to be used to compare the different backends for my master thesis :)

mrobakowski:
I want to do a PR with the experiments later, when they're fully done and there's more of them. But this infrastructure stuff can be merged independently of that and could prove useful for some other benchmarks and such.

Max-Meldrum:
Rather than having 2 features. I would suggest adding x86 dependency as following:

```bash
[target.'cfg(target_arch = "x86")'.dependencies]
x86 = "version", optional?
```

and then turn the measure function into something like this:

```rust
 pub fn measure<T>(operation: &'static str, func: impl FnOnce() -> T) -> (T, DataPoint) {
   #[cfg(any(target_arch = "x86"))]
   {
     // rdtsc...
   }
  #[cfg(not(any(target_arch = "x86")))]
  {
    // other..
  } 
 }
```

Or was the point to be able to use the 2 different measures?

mrobakowski:
I wanted to have both the `rdtsc` way and the `Instant` one since the latter gives you actual units of time. `rdtsc` gives you _something_ that monotonically increases based on the processor's mood, but it is ultra-low-cost. The measure fns are mutually exclusive and I put them in modules so I can have some additional global state for the `Instant`-using version. I guess I could just have one fn and then put another `#[cfg(...)]` on the static variable.

Max-Meldrum:
Okay, well if that is the case, then it is fine. But what about the x86 dependency.  if it is not under a [target.'cfg(target_arch = "x86")'.dependencies], what will happen if the arch is not x86 and you try to run the rdtsc feature?

mrobakowski:
You'll get an error, as you should. I wouldn't want it to silently switch to using `Instant`s instead.

Max-Meldrum:
Right. I keep forgetting that this is purely for testing/benching. Nice work! Ready to be merged?

mrobakowski:
I think so

Max-Meldrum:
After some additional thinking, I still think it should be put under target_arch dependency and tie the feature to the archs. As although our CI is likely to run on ``x86``, we wouldn't want the compilation to fail when running ``cargo build --all-features`` on some other arch.
```rust
[target.'cfg(target_arch = "x86")'.dependencies]
```

```rust
#[cfg(any(target_arch = "x86", target_arch = "x86_64"), feature = "metered_state_backend_rdtsc")]
```

mrobakowski:
Fair enough

Issue #94, opened 2020-04-20
Non-unique serialisation ids


Bathtor:
```rust
impl<A: ArconType> UnsafeSerde<A> {
    const SID: kompact::prelude::SerId = 26;
}
```
This is a really bad idea. Assigning the same serialisation id to multiple concrete serialisers (i.e. one for every generic type) can end you up with someone trying to deserialise data of type B with a serialiser for  type A and getting *some* weird as shit value by accident without error.

This may work ok for protobuf stuff assuming it maintains its own internal ids (which I'm not sure it does, tbh), but for abomonation this is definitely a no-no.

It would be better to generate serialisation id for every type `A` and have `ReliableSerde<A>` use `A::SER_ID` instead, so that it can't be confused with `B`.

Kompact has the following trait for this purpose to be implemented on `A` *not* `ReliableSerde<A>`:
```rust
pub trait SerialisationId {
    const SER_ID: SerId;
}
```
Alternatively you can make add a `SER_ID` field to `ArconType`. 

We will need a way to generate those ids reliably and consistently from the Arc compiler toolchain at some point.

Max-Meldrum:
yeah, good point. I guess the current approach works under the assumption that the system is always set up correctly ^^

I'm wondering if we might need a version identifier in addition to a SER_ID.  First of all, we need to know whether if it is Protobuf or Abomonation. Secondly, it would be good to check that their versions match.

If we don't have predefined identifiers for Protobuf/Abomonation, how  would the following logic work in the Arcon node? 

```rust
fn receive_network(&mut self, msg: NetMessage) {
        let arcon_msg: ArconResult<ArconMessage<IN>> = match *msg.ser_id() {
            ReliableSerde::<IN>::SER_ID => msg
                .try_deserialise::<ArconMessage<IN>, ReliableSerde<IN>>()
                .map_err(|_| arcon_err_kind!("Failed to unpack reliable ArconMessage")),
            UnsafeSerde::<IN>::SER_ID => msg
                .try_deserialise::<ArconMessage<IN>, UnsafeSerde<IN>>()
                .map_err(|_| arcon_err_kind!("Failed to unpack unreliable ArconMessage")),
            _ => panic!("Unexpected deserialiser"),
        };
}
```

Would ``IN: ArconType`` have to impl ``UNSAFE_SER_ID``,  ``RELIABLE_SER_ID``  and match on ``msg.ser_id()``

Bathtor:
[...]

PR #101, opened 2020-04-28
[WIP] improve flight serde

Max-Meldrum:
Addresses #94 

ArconType has a new look:

```rust
pub trait ArconType: ArconTypeBounds
where
    Self: std::marker::Sized,
{
    /// Serialisation ID for Arcon's Unsafe In-flight serde
    const UNSAFE_SER_ID: SerId;
    /// Serialisation ID for Arcon's Reliable In-flight serde
    const RELIABLE_SER_ID: SerId;
    /// Current version of this ArconType
    const VERSION_ID: VersionId;
}
```
There is now a single arcon proc macro.

```rust
#[arcon(unsafe_ser_id = 104, reliable_ser_id = 105, version = 1)]
pub struct Data {
   #[prost(uint32, tag = "1")]
    pub id: u32,
}
// or
#[arcon(unsafe_ser_id = 104, reliable_ser_id = 105, version = 1, keys = id)]
pub struct Data {
   #[prost(uint32, tag = "1")]
    pub id: u32,
}
```

Flight Serde now adds the version identifier to the start of the serialised data and checks this at deserialisation:

```rust
impl<A: ArconType> Deserialiser<RawArconMessage<A>> for ReliableSerde<A> {
     const SER_ID: SerId = A::RELIABLE_SER_ID;

     fn deserialise(buf: &mut dyn Buf) -> Result<RawArconMessage<A>, SerError> {
          let version_id = buf.get_u32();
          if version_id != A::VERSION_ID {
              let err = format!(
                  "Mismatch on ArconType version. Got {} while expecting {}",
                  version_id,
                  A::VERSION_ID
              );
              return Err(SerError::InvalidData(err));
          }

          let arcon_message = RawArconMessage::decode(buf.bytes()).map_err(|_| {
              SerError::InvalidData("Failed to decode RawArconMessage".to_string())
          })?;
          Ok(arcon_message)
     }
}
```
Thoughts?

Added a WIP tag to the PR as I want to still fix the flight serde benches.

### DIFF
--- a/execution-plane/arcon/arcon_macros/src/lib.rs
+++ b/execution-plane/arcon/arcon_macros/src/lib.rs
@@ -13,30 +13,110 @@ use proc_macro::TokenStream;
 use proc_macro2::{Ident, Span};
 use syn::{parse_macro_input, DeriveInput};
 
-/// arcon is a macro that helps define a Arcon supported struct
+const VERSION_ID: &str = "version";
+const UNSAFE_SER_ID: &str = "unsafe_ser_id";
+const RELIABLE_SER_ID: &str = "reliable_ser_id";
+const KEYS: &str = "keys";
+
+/// arcon is a proc macro for defining an ArconType struct
 ///
-/// By default, the macro will implement [std::hash::Hasher] that hashes on all fields of the
-/// struct. Use [arcon_keyed] instead if the struct needs to be hashed on specific fields.
+/// #[arcon(reliable_ser_id = 1, unsafe_ser_id = 2, version = 1)]
+/// or
+/// #[arcon(reliable_ser_id = 1, unsafe_ser_id = 2, version = 1, keys = id)]
 #[proc_macro_attribute]
-pub fn arcon(metadata: TokenStream, input: TokenStream) -> TokenStream {
+pub fn arcon(args: TokenStream, input: TokenStream) -> TokenStream {
     let item = parse_macro_input!(input as DeriveInput);
     let name = &item.ident;
-    let _ = proc_macro2::TokenStream::from(metadata);
+    let arcon_args: Vec<(String, String)> = args
+        .to_string()
+        .trim()
+        .split(',')
+        .map(|s| {
+            let a: Vec<String> = s
+                .trim()
+                .to_string()
+                .split(" ")
+                .map(|s| s.to_string())
+                .collect();
+            assert_eq!(a.len(), 3, "Expecting the following arg setup: unsafe_ser_id = 1, reliable_ser_id = 2, version = 1");
+            assert_eq!(a[1], "=");
+            (a[0].to_string(), a[2].to_string())
+        })
+        .filter(|(arg_name, _)| {
+            arg_name == VERSION_ID
+                || arg_name == UNSAFE_SER_ID
+                || arg_name == RELIABLE_SER_ID
+                || arg_name == KEYS
+        })
+        .collect();
+
+    assert!(
+        arcon_args.len() >= 3 && arcon_args.len() <= 4,
+        "arcon requires at least 3 args: unsafe_ser_id, reliable_ser_id, version"
+    );
+
+    let raw_ids: Vec<u64> = arcon_args
+        .iter()
+        .filter(|(arg_name, _)| arg_name != KEYS && arg_name != VERSION_ID)
+        .map(|(_, arg_value)| arg_value.parse::<u64>().unwrap())
+        .collect();
+
+    assert_eq!(raw_ids.len(), 2);
+    assert_ne!(raw_ids[0], raw_ids[1], "UNSAFE_SER_ID and RELIABLE_SER_ID must have different values");
+
+    let ids: Vec<proc_macro2::TokenStream> = arcon_args
+        .iter()
+        .filter(|(arg_name, _)| arg_name != KEYS)
+        .map(|(arg_name, arg_value)| {
+            if arg_name == RELIABLE_SER_ID {
+                let value = arg_value.parse::<u64>().unwrap();
+                quote! { const RELIABLE_SER_ID: SerId  = #value; }
+            } else if arg_name == UNSAFE_SER_ID {
+                let value = arg_value.parse::<u64>().unwrap();
+                quote! { const UNSAFE_SER_ID: SerId = #value; }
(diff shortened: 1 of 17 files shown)


Bathtor:
> Flight Serde now adds the version identifier to the start of the serialised data and checks this at deserialisation

Is this duplicate with what protocol buffers do internally? Or are they not doing this? I can't remember.

Max-Meldrum:
> > Flight Serde now adds the version identifier to the start of the serialised data and checks this at deserialisation
> 
> Is this duplicate with what protocol buffers do internally? Or are they not doing this? I can't remember.

Not sure how their schema evolution is implemented. But, I would say it is not really a duplicate. We are just checking our flight serde. That the sender and receiver are on the same version.  

```rust
#[arcon(unsafe_ser_id = 104, reliable_ser_id = 105, version = 1)]
pub struct Data {
   #[prost(uint32, tag = "1")]
    pub id: u32,
    #[prost(uint32, tag = "2")]
    pub price: u32,
}
// new
#[arcon(unsafe_ser_id = 104, reliable_ser_id = 105, version = 2)]
pub struct Data {
   #[prost(uint32, tag = "1")]
    pub id: u32,
}
```
Although the serde between version 1 and 2 here might work, it shouldn't really occur as both (sender/receiver) should be on the same version.

Bathtor:
> We are just checking our flight serde. That the sender and receiver are on the same version.

But is that what we want? Like this, if we do reconfiguration we gotta replace every component at once. With protobuf's version update mechanism we could replace components one by one from downstream to upstream (since downstream components will be able to handle the older version messages from upstream components). Of course this mechanism immediately fails with loops, which clearly have to be replaced all at once. But anyway...something to consider perhaps?

Max-Meldrum:
ah, right. That is true.  yeah, probably worth to think about 😄

mrobakowski:
I still hate the manual `prost` annotation :P. You could fully infer them from the field types and order (which `serde` does, unlike `prost`...). `prost` is targeted at cases where a protobuf definition file is the source of truth. `serde` - where the rust code is the source of truth. So IMO we should either generate the structs from protobuf definition files, or just use `serde` instead of `prost`.

mrobakowski:
Also, I'd prefer if the syntax for the macro was just `#[derive(ArconType)]`.

Max-Meldrum:
> I still hate the manual `prost` annotation :P. You could fully infer them from the field types and order (which `serde` does, unlike `prost`...). `prost` is targeted at cases where a protobuf definition file is the source of truth. `serde` - where the rust code is the source of truth. So IMO we should either generate the structs from protobuf definition files, or just use `serde` instead of `prost`.

I'm not a fan of the prost annotation either. But it is important to remember that the code is supposed to be generated from Arc, that is, only us Arcon developers will see the Rust code.  While it is good to make things more pretty, it has to be worth the time :)

Although we have not tried it yet, the plan is to indeed generate the structs from a protobuf file that has been generated by the Arc compiler. But having the option to declare protobuf based structs in Rust directly is nice for development/testing.

While prost is not a hard requirement for arcon, Protobuf is.  Using just  `serde` will not provide us with schema evolution, cross system portability (i.e. export/import arcon data into another system that is not Rust lang).

Max-Meldrum:
> Also, I'd prefer if the syntax for the macro was just `#[derive(ArconType)]`.

I believe I tried using that approach, but it was not possible to define other #[derive] inside a derive itself.

```rust
 #[derive(Clone, ::abomonation_derive::Abomonation, ::prost::Message)]
 #item
```

But if you have suggestions on how to improve this, would be cool to hear :)

mrobakowski:
We can go the `std` route and just require the user (arc compiler, I guess) to just write all the derives at the top level. Standard library does that. For example:
```rust
#[derive(Copy)]
struct Foo;
```
fails to compile with error
```
error[E0277]: the trait bound `Foo: std::clone::Clone` is not satisfied
 --> src/lib.rs:1:10
  |
1 | #[derive(Copy)]
  |          ^^^^ the trait `std::clone::Clone` is not implemented for `Foo`
  |
  = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

Max-Meldrum:
Yeah, perhaps it is better to go that way and have the following in the prost generation.
```rust
let mut config = prost_build::Config::new();
config.type_attribute(".", "#[derive(Clone, serde::Serialize, serde::Deserialize, arcon::ArconType)]");
```

mrobakowski:
For additional configuration most libraries do
```rust
#[derive(Foo)]
#[foo(some_struct_wide_config = "bar")]
struct Baz {
    #[foo(some_field_config = "qux")]
    quux: String
}
```

Max-Meldrum:
Removed the version check for Protobuf and tested some setups.

The Sender always sends ``ArconDataTest`` and the Receiver tries to Deserialise into either ``RemovedField`` or ``AddedField``.  While I assumed it would manage to deserialise RemovedField, it also deserialises AddedField with no problems (probably outputting an empty ``bytes`` vec).

```rust
     // Sending data
    #[arcon(unsafe_ser_id = 104, reliable_ser_id = 105, version = 1)]
    pub struct ArconDataTest {
        #[prost(uint32, tag = "1")]
        pub id: u32,
        #[prost(uint32, repeated, tag = "2")]
        pub items: Vec<u32>,
        #[prost(uint32, tag = "3")]
        pub price: u32,
    }
    
    // Receiving Data
    #[arcon(unsafe_ser_id = 104, reliable_ser_id = 105, version = 1)]
    pub struct RemovedField {
        #[prost(uint32, tag = "1")]
        pub id: u32,
        #[prost(uint32, repeated, tag = "2")]
        pub items: Vec<u32>,
    }

    #[arcon(unsafe_ser_id = 104, reliable_ser_id = 105, version = 1)]
    pub struct AddedField {
        #[prost(uint32, tag = "1")]
        pub id: u32,
        #[prost(uint32, repeated, tag = "2")]
        pub items: Vec<u32>,
        #[prost(uint32, tag = "3")]
        pub price: u32,
        #[prost(bytes, tag = "4")]
        pub bytes: Vec<u8>,
    }
```

Max-Meldrum:
Merging this. Leaving bench fixes for #103

Issue #75, opened 2020-02-22
Key extraction


Max-Meldrum:
Rather than forcing Hash upon all ArconType's, look into having our own trait for extracting a key. That is, we don't only want to be able to hash keys, users may want their own custom logic for defining keys.

Max-Meldrum:
```rust
// v1
trait ArconKey {
  get_key(&self) -> u64; // return 0 if no key?
}
// v2
trait ArconKey {
  get_key(&self) -> Option<u64>;
}

// if we enforce ArconKey on the ArconTypeBounds, then there are no issues.
// but, if we would like something like KeyBy<A: ArconType + ArconKey>, 
// then we would have to enforce it for all channel strategies with the way it is implemented currently.
pub enum ChannelStrategy<A>
where
    A: ArconType + ArconKey,
{
    Forward(forward::Forward<A>),
    KeyBy(key_by::KeyBy<A>), 
}
```

Bathtor:
I think things should *always* have a key so I prefer v1. We can always fall back on a `derive(ArconKey)` based on a hash of the whole struct if no custom logic is needed, and if the strategy (like forward) doesn't need it, it will never be called, so it doesn't matter if it's unnecessarily expensive.

mrobakowski:
I live v1 better. Either a type implements the trait and provides a sensible result or just doesn't. We don't really have to worry about the "no key" scenario because such types just shouldn't impl the trait.

... but I don't know how to do this while keeping `ChannelStrategy` an enum. If it was a trait, we could do something like:
```rust
struct Forward<A>;
impl<A: ArconType> ChannelStrategy<A> for Forward<A> {}

struct KeyBy<A>;
impl<A: ArconType + ArconKey> ChannelStrategy<A> for KeyBy<A> {}
```

Also, I don't really like the name `ArconKey`. `ArconKeyable` or just `HasKey` reflects what it does better.

Max-Meldrum:
> I think things should _always_ have a key so I prefer v1. We can always fall back on a `derive(ArconKey)` based on a hash of the whole struct if no custom logic is needed, and if the strategy (like forward) doesn't need it, it will never be called, so it doesn't matter if it's unnecessarily expensive.

Was leaning towards this approach as well. But then there is no point in having a trait for it I guess?

```rust
pub trait ArconType: ArconTypeBounds
where
    Self: std::marker::Sized,
{
    /// Serialisation ID for Arcon's Unsafe In-flight serde
[...]

PR #104, opened 2020-05-02
Arcon derive + Proto Gen

### LINKED ISSUE #102: Arcon derive and Protobuf generation
From discussions in #101 

Look to turn #[arcon] into a derive macro instead and add test cases for generating ArconType structs from a Protobuf file.

```rust
#[derive(ArconType)] or #[derive(Arcon)]
#[arcon(unsafe_ser_id = 104, reliable_ser_id = 105, version = 1)]
pub struct Data {
   pub id: u32,
}
```

Where proto generation might look like this:

```rust
let mut config = prost_build::Config::new();
config.type_attribute(".", "#[derive(Clone, arcon::Arcon)]");
let arcon_cfg = format!("#[arcon(unsafe_ser_id = {}, reliable_ser_id = {}, version = {})]",  id, id, version);
config.type_attribute(".",  arcon_cfg);
config.compile_protos(&["file.proto"], &["proto_dir/"]).unwrap();
```

Max-Meldrum:
This PR adds the following (closes #102):

1. an Arcon derive macro that is more elegant and less hacky than its predecessor.
2.  ``proto_tests`` crate that verifies Protobuf files can be generated into valid ArconType's.

For number 2, it is at the moment not possible to generate Protobuf messages with ``oneof's`` while not specifying any specific ``key``. This is due to the ``oneof`` being an Option and it will therefore not implement ``Hash``. However, this should be resolved when we fix key extraction (#75).

### DIFF
--- a/execution-plane/Cargo.lock
+++ b/execution-plane/Cargo.lock
@@ -13,7 +13,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e50e2a046af56a864c62d97b7153fda72c596e646be1b0c7963736821f6e1efa"
 dependencies = [
  "proc-macro2 1.0.10",
- "quote 1.0.3",
+ "quote 1.0.4",
  "synstructure",
 ]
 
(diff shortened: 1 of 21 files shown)


Max-Meldrum:
@mrobakowski Now the macro looks better. Any change suggestion or shall I merge?

mrobakowski:
LGTM :+1:

PR #106, opened 2020-05-06
add get_key method to ArconType and proto doc derive support

Max-Meldrum:
Adds a get key method on ArconType:

```rust
pub trait ArconType: ArconTypeBounds
where
    Self: std::marker::Sized,
{
    /// Return the key of this ArconType
    fn get_key(&self) -> u64;
}
```

The struct down below:
```rust
#[derive(Arcon, prost::Message, Clone, abomonation_derive::Abomonation)]
#[arcon(unsafe_ser_id = 104, reliable_ser_id = 105, version = 1, keys = "id")]
pub struct Item {
   #[prost(uint64, tag = "1")]
   pub id: u64,
   #[prost(uint32, tag = "2")]
   pub price: u32,
}
```

will implement ``get_key`` as following:

```rust
fn get_key(&self) -> u64 {
   let mut state = ::std::collections::hash_map::DefaultHasher::new();
   self.id.hash(&mut state);
   state.finish()
}
```

It now also works to derive the arcon attributes from the Protobuf files (if the #[arcon(..)] attr is not found, it will attempt to parse them from the doc comments). The proto definition below will implement the same get_key function as above.
```proto
// unsafe_ser_id = 104
// reliable_ser_id = 105
// version = 1
// keys = id
message Item {
  uint64 id = 1;
  uint32 price = 2;
}
```

### DIFF
--- a/execution-plane/arcon/arcon_macros/src/lib.rs
+++ b/execution-plane/arcon/arcon_macros/src/lib.rs
@@ -26,10 +26,8 @@ pub fn arcon(input: TokenStream) -> TokenStream {
     let input: DeriveInput = syn::parse(input).unwrap();
     let name = &input.ident;
 
-    let meta = input
-        .attrs
-        .iter()
-        .find_map(|attr| match attr.parse_meta() {
+    let (unsafe_ser_id, reliable_ser_id, version, keys) = {
+        let arcon_attr = input.attrs.iter().find_map(|attr| match attr.parse_meta() {
             Ok(m) => {
                 if m.path().is_ident("arcon") {
                     Some(m)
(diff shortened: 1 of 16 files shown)


Bathtor:
Should we try (at some point) to maybe pick a "good" hasher depending on the fields to be hashed? For example for uint64 SIP will be slow, but have good statistical properties, while something like fxhash may be very fast but with bad properties. But more importantly, for structs where all fields are part of the hash and some of them may be big (byte arrays, strings...) things likes XXHash may be a much better idea.

Max-Meldrum:
Yeah, that is a good idea. I used the DefaultHasher in the macro as it is easy to "import". Should this selection happen above the derive macro? i.e. add another attribute ``hasher``?

Bathtor:
> Yeah, that is a good idea. I used the DefaultHasher in the macro as it is easy to "import". Should this selection happen above the derive macro? i.e. add another attribute `hasher`?

I'm actually thinking we could do this smartly as part of the derive macro using some heuristics? Basically look at all the keys that have been selected for the key, produce an estimate of the number of bytes to be hashed every time `get_key` is invoked and pick a hasher based on this estimate.

If you check the Rust crate for [XXHash](https://github.com/shepmaster/twox-hash) they have some pretty nice statistics already on where the cut-off points likely are, which we could use as a starting point.

For the estimated number of bytes heuristic we can start out with something simple, like assume every string or byte array has at least 10 characters/bytes. And for other data types we know the number of bytes, of course.

Max-Meldrum:
Okay, that sounds cool. Just wondering, could there be potential problems during schema evolution? Perhaps the Hasher is now changed and suddenly data is sent across the network differently.

Bathtor:
> Okay, that sounds cool. Just wondering, could there be potential problems during schema evolution? Perhaps the Hasher is now changed and suddenly data is sent across the network differently.

I mean, the same is true if the (set of) keys change, in general. And if the keys don't change, hasher shouldn't change. So...I don't see an issue right now.

Max-Meldrum:
Perhaps someone knows the answer to this.  Currently, in the macro, it looks like this:

```rust
quote! { const RELIABLE_SER_ID: SerId  = #reliable_ser_id; }
```

The only reason it does not look like the following:

```rust
quote! { const RELIABLE_SER_ID: ::arcon::SerId  = #reliable_ser_id; }
```

is that I get issues when using the derive macro within the arcon crate itself (for tests). "Cannot find {{root}} or something.

It would simplify imports and stuff.

```rust
// arcon
pub use fxhash::FxHasher;
pub use xxhash::XXHash;

// macro
let mut state = ::arcon::FxHasher::new();
```

Bathtor:
I think the `$crate::X` should be doing the right thing here, no?

Max-Meldrum:
> I think the `$crate::X` should be doing the right thing here, no?

I tried it but didn't work. I googled a bit and seems like it is a general issue. $crate is not supported in procedural macros.

Bathtor:
> > I think the `$crate::X` should be doing the right thing here, no?
> 
> I tried it but didn't work. I googled a bit and seems like it is a general issue. $crate is not supported in procedural macros.

Well...that's a nuisance. I thought they fixed that -.-

Bathtor:
https://docs.rs/quote/0.6.13/quote/macro.quote_spanned.html

This would work if arcon would be available in the macro...but it isn't, right? It's like Kompact, where core depends on the macros, but not the other way around, right?

Max-Meldrum:
yeah, ``arcon`` depends on ``arcon_macros``.

It actually worked doing this:

```rust
// arcon src/lib.rs
extern crate self as arcon;

// macro
quote! { const RELIABLE_SER_ID: arcon::SerId  = #reliable_ser_id; }
```

Bathtor:
> yeah, `arcon` depends on `arcon_macros`.
> 
> It actually worked doing this:
> 
> ```rust
> // arcon src/lib.rs
> extern crate self as arcon;
> 
> // macro
> quote! { const RELIABLE_SER_ID: arcon::SerId  = #reliable_ser_id; }
> ```

But since that isn't from the root, technically it could resolve to the wrong arcon module if someone has another "arcon"-module in scope at the call-site of the macro, no?

Max-Meldrum:
Hmm, not sure I understand. You mean if there is some crate ``arcon_something`` that depends on ``arcon`` but is then used in ``arcon_macros``? 

```rust
// arcon_macros
arcon_something::somethingsomething
```

Bathtor:
> Hmm, not sure I understand. You mean if there is some crate `arcon_something` that depends on `arcon` but is then used in `arcon_macros`?
> 
> ```rust
> // arcon_macros
> arcon_something::somethingsomething
> ```

Nah, I was thinking if there's a crate `x` using arcon macros that has a module called `arcon` which is in scope somewhere instead of the `arcon` crates' root module, your macros can resolve to `x::arcon::SerId` instead of `::arcon::SerId`, no?

Max-Meldrum:
> > Hmm, not sure I understand. You mean if there is some crate `arcon_something` that depends on `arcon` but is then used in `arcon_macros`?
> > ```rust
> > // arcon_macros
> > arcon_something::somethingsomething
> > ```
> 
> Nah, I was thinking if there's a crate `x` using arcon macros that has a module called `arcon` which is in scope somewhere instead of the `arcon` crates' root module, your macros can resolve to `x::arcon::SerId` instead of `::arcon::SerId`, no?
> 
> It's admittedly a very silly edge case^^

ah, I see. Well, can't see that ``arcon_macros`` would be used from anywhere else than ``arcon`` (re-export).

Bathtor:
> ah, I see. Well, can't see that `arcon_macros` would be used from anywhere else than `arcon` (re-export).

I think that may also happen if someone uses the forwarded macros from the arcon crate, since the type resolution happens at call-site by default, no?

Max-Meldrum:
Okay, I think I understand it better now. I'll add an arcon module in the experiments crate and check what will happen.

Max-Meldrum:
Added hasher selection on estimated bytes using some advanced AI techniques. Might be good later to test how well they distribute etc.. And see if MurmurHash has okay properties while still being good for smaller values.

```rust
fn hasher_stream(estimated_bytes: usize) -> proc_macro2::TokenStream {
    // Selection Process:
    //
    // Low bytes estimation: Pick a hasher suited for small values
    // Medium bytes estimation: Pick Rust's default hasher
    // High bytes estimation: Pick XXHash that performs much better on larger values

    if estimated_bytes <= 8 {
        quote! { ::arcon::FxHasher::default(); }
    } else if estimated_bytes <= 32 {
        quote! { ::std::collections::hash_map::DefaultHasher::new(); }
    } else {
        quote! { ::arcon::XxHash64::default(); }
    }
}
```

execution-plane/arcon/arcon_macros/src/lib.rs
```diff
@@ -38,14 +50,218 @@ pub fn arcon(input: TokenStream) -> TokenStream {
...
+    let mut bytes = 0;
+    for field in s.fields.iter() {
+        match field.ident {
+            Some(ref ident) => {
+                if fields.contains(&ident.to_string()) {
+                    if let syn::Type::Path(p) = &field.ty {
+                        for segment in &p.path.segments {
+                            let ref ident = segment.ident;
+                            let cleaned = ident.to_string();
+                            if cleaned == STRING_IDENT {
```
mrobakowski, line 193: a match would be nicer here

Max-Meldrum:
I'll merge this if there are no other comments. 

Regarding the macro resolve thing,  yeah if I add an arcon module to ``experiments/nexmark/``, then it fails to find ``arcon::Arcon``, but as the arcon crate is not meant to be used directly and that we will control the codegen part. I don't think it's worth spending time on this. But if someone wants to have a look at it, feel free :)

PR #110, opened 2020-05-18
State Bundles

mrobakowski:
This PR does several things, unfortunately. I'm afraid I'm unable to split it into several small self-contained PRs.

* extract state management to a separate crate: `arcon_state` (more on that later)
* make `arcon` work with `arcon_state`
* make all functional operators state-aware
* add some state backend benchmarks to `experiments/bin/perf.rs`
* use `structopt` instead of `clap` in `experiments`

The most significant part is the new `arcon_state` crate. It's a redesign of the old `state_backend` module that removes some excessive generic-ness and introduces several new concepts that increase safety and make the API nicer.

I simplified and unified the requirements for types that can be persisted in a state backend. There's no optional dependency on `serde` anymore (sans what is needed for integration with `faster-rs`). The new crate now always depends on protobuf serialization. Because of those unified requirements, the whole complicated state builder system was removed. No more weird dynamic builders generated by complicated macros.

State `Bundle`s are a big addition. Here's a taste:
```rust
arcon_state::bundle! {
    /// My test bundle
    pub struct MyBundle<T: Value> {
        /// some value
        value: Handle<ValueState<T>>,
        /// some map
        some_special_map: Handle<MapState<String, f64>, i32, u32>
    }
}
impl<T: Value> MyBundle<T> {
    fn new() -> Self {
        MyTestBundle {
            value: Handle::value("value"),
            some_special_map: Handle::map("special_map").with_item_key(-1).with_namespace(0)
        }       
    }   
}

// Bundles are always registered before usage - usually by a `Node`. Unsafe is here just to 
// signal that this shouldn't be called from wherever.
let mut bundle = MyBundle::<u32>::new();
bundle.register_states(&mut unsafe { RegistrationToken::new(&mut backend_session) });

let mut active_bundle = bundle.activate(&mut backend_session);
let mut value_handle = active_bundle.value();
value_handle.set(3).unwrap();
println!("value is {:?}", value_handle.get().unwrap());
let mut map_handle = active_bundle.some_special_map();
map_handle.insert("foo".into(), 0.5).unwrap();
println!("map[foo] is {:?}", map_handle.get(&"foo".into()).unwrap())
```
`Bundle`s are collections of (inactive) `Handle`s. `Bundle`s can be activated by passing a state session to them, which replaces passing a state backend reference in most places. When activated, the bundle is turned into a special active bundle type generated by the `bundle!` macro. This type implements methods for getting active handles to declared states. Crucialy, the bundle can be activated only after it has been registered. This saves us from the problems with RocksDb, which requires you to state all the column families before opening the database (we use a separate column family per state).

A lot of the code has been made generic and a lot of stuff got unboxed. This has the potential for allowing more compiler optimizations. The downside is that some code got more complex (more generic parameters to keep track of), but I believe this could be fixed in the future by more liberal usage of `impl Trait` in argument positions.

Functional operators, such as `FlatMap`, can be stateful now:
```rust
let not_stateful = FlatMap::new(|x| x+5);
let stateful = FlatMap::stateful(
    MyBundle::<u32>::new(),
    |x, state, session| {
        let mut active = state.activate(session);
        let res = x + active.value().get().unwrap().unwrap();
        active.value().set(x).unwrap();
        res
    }
);
```


mrobakowski:
Most of the concrete-backend-specific code is adopted from the old `state_backen` module, but git didn't really notice that and that's where the huge diff is coming from.

Max-Meldrum:
Cool! Nice work :) 

You mentioned unboxing and potential compiler optimisations. Did you notice any difference in performance when using the Metered stuff now?

mrobakowski:
The performance stayed roughly the same. I'll have to profile it and maybe add some `#[inline]` annotations.

segeljakt:
Up here from above this looks like just what we need to implement any single-input combinator

Is there a possibility to support 2 inputs later on for the capability of joins? like Flink's [CoProcessFunction](https://ci.apache.org/projects/flink/flink-docs-stable/dev/stream/operators/process_function.html)?

mrobakowski:
@segeljakt you're right. We can actually even do without `Map` or `Filter` because they can be trivially implemented via `FlatMap`. 

I think doing a join manually and processing pairs should be doable with stateful flatmap even right now, albeit hacky. The two inputs have to be unified to a single type with something like `Map::new(Either::Left)` and `Map::new(Either::Right)` respectively and then these two nodes can be connected to a single stateful node which can do bookkeeping. Something like this is done in [NEXMark query \#3](https://github.com/mrobakowski/arcon/tree/experiment/state-bundles/execution-plane/arcon/experiments/src/nexmark/queries/q3.rs) (around line 55 and around line 160). Actually, in that case, there's still a single source node, but doing it with multiple sources is just a matter of setting multiple `in_channels` for the node that is doing bookkeeping.

Some common case version of this should definitely be abstracted away.

segeljakt:
I don't know fully enough how Arcon works yet but maybe it can be beneficial to have intrinsics for alleviating some overhead or make scheduling better. On Arc's side it does not change much I think.

mrobakowski:
CI is having some troubles with MacOS. I don't have a mac, so I don't know how to fix this :(

Max-Meldrum:
> CI is having some troubles with MacOS. I don't have a mac, so I don't know how to fix this :(

Looks to me like it is trying to compile libfaster on osx (which it shouldn't).

mrobakowski:
There is quite a big regression with Faster state backend. Well, it is not exactly a regression, because the code was just incorrect before. `SourceContext`s never created state backend sessions and now they do it per processed item, so we get stupidly short sessions in many cases. 

There are two solutions to this: a quick and dirty one is to just add some more (`Ref`)`Cell`s inside the `SourceContext`, create the session outside tight loops, and pass it to the `process` method; the more proper solution would be to re-arrange `SourceContext` to be able to borrow parts of it independently. I lean towards the second solution as it would also benefit `Node`, which has a similar problem and that would remove a bunch of `RefCell`s from it. 

But I think the second solution should be separated into another (way smaller) PR.

execution-plane/arcon/arcon_state/src/faster/aggregator_ops.rs
```diff
@@ -0,0 +1,106 @@
...
+        serialized.push(VALUE_MARKER);
+        protobuf::serialize_into(&mut serialized, &value)?;
+
+        // See the make_aggregate_fn function in this module. Its result is set as the
+        // merging operator for this state.
+        self.aggregate(&key, serialized, handle.id)
+    }
+}
+
+pub fn make_aggregate_fn<A>(aggregator: A) -> Box<dyn Fn(&[u8], &[u8]) -> Vec<u8> + Send>
```
Max-Meldrum, line 56: Quite many  ``.except()`` cases in this function. Perhaps the return should be a Result<>?

execution-plane/arcon/arcon_state/src/faster/reducer_ops.rs
```diff
@@ -0,0 +1,54 @@
...
+        value: T,
+    ) -> Result<()> {
+        let key = handle.serialize_id_and_metakeys()?;
+        let serialized = protobuf::serialize(&value)?;
+
+        self.aggregate(&key, serialized, handle.id)
+    }
+}
+
+pub fn make_reduce_fn<T, F>(reducer: F) -> Box<dyn Fn(&[u8], &[u8]) -> Vec<u8> + Send>
```
Max-Meldrum, line 41: Same here, return a Result<> and avoid ``excepts``?

execution-plane/arcon/arcon_state/src/lib.rs
```diff
@@ -0,0 +1,364 @@
+// Copyright (c) 2020, KTH Royal Institute of Technology.
+// SPDX-License-Identifier: AGPL-3.0-only
+#![feature(associated_type_defaults)]
```
Max-Meldrum, line 3: Are these features gonna hit stable anytime soon?

execution-plane/arcon/arcon_state/src/rocks/vec_ops.rs
```diff
@@ -0,0 +1,229 @@
...
+        // cannot really consume the `values` iterator twice, so just preallocate a bunch of bytes
+        let mut serialized = Vec::with_capacity(256);
+
+        // reserve space for the length
+        fixed_bytes::serialize_into(&mut serialized, &0usize)?;
+        let mut len = 0usize;
+
+        for elem in values {
+            len += 1;
+            protobuf::serialize_into(&mut serialized, &elem)?;
```
Max-Meldrum, line 138: What will happen if the amount of bytes go above 256 here in this loop? does the ``protobuf::serialize_into```allocate more memory into the memory block?

mrobakowski: Yes, `BufMut` impl for `Vec<u8>` just grows the vec if the capacity is reached.

execution-plane/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 members = [
   "arcon",
   "arcon/experiments",
+  "arcon/arcon_state",
```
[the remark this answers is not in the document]
mrobakowski: I don't know how that would interact with testing on CI. We definitely want the tests to be run

execution-plane/arcon/arcon_state/src/lib.rs
mrobakowski: https://github.com/rust-lang/rust/issues/29661 no idea. We don't necessarily need this, it just makes the code slightly shorter. We're not targeting stable anyway, so I see no good reason to remove this. But if we wanted to, it's easy to remove.

execution-plane/arcon/arcon_state/src/faster/aggregator_ops.rs
mrobakowski: It's not very easy to do, because the expects are inside the closure that is returned. Hmm... I'd have to add an error-carrying variant to the types that implement `FasterRmw`, where the closures ultimately get used. That's probably a better solution, because panicking across ffi boundaries is UB as far as I know.

execution-plane/Cargo.toml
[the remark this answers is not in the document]
Max-Meldrum: The CI runs ``cargo test --all``, so it runs tests also in the crates under arcon/.  (arcon_macros etc..)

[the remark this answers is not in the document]
mrobakowski: According to the cargo book, `--all` is a deprecated alias to `--workspace`, which runs tests from all the members of the workspace.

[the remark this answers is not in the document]
mrobakowski: According to the cargo book, `--all` is a deprecated alias to `--workspace`, which runs tests from all the members of the workspace.

execution-plane/arcon/arcon_state/src/lib.rs
Max-Meldrum: While we don't exactly need to run on stable right now, I think it would be nice to aim towards this in the future, so that is why I am asking. Nightly features are fine, but would like to avoid to rely  on features that may take 2-3 years to get merged.

execution-plane/Cargo.toml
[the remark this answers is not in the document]
Max-Meldrum: okay, well, try run that command in the execution-plane workspace and you will see it runs tests under arcon_macros for example.

execution-plane/arcon/arcon_state/src/faster/aggregator_ops.rs
mrobakowski: Fixed in 8cf8d39

execution-plane/arcon/arcon_state/src/faster/reducer_ops.rs
mrobakowski: Fixed in 8cf8d39

execution-plane/arcon/arcon_state/src/faster/aggregator_ops.rs
```diff
@@ -0,0 +1,106 @@
+use crate::{
```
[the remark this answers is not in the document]
mrobakowski: Fixed in 4905f1f. I hope I didn't miss anything.

execution-plane/Cargo.toml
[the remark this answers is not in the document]
mrobakowski: Fixed in 87dc819. You're right, the tests seem to still run.

execution-plane/arcon/arcon_state/src/lib.rs
mrobakowski: Oh, the comment I left here is just for `associated_type_defaults`. The `const_generics` feature is to be able to use fixed-length arrays as item keys or namespaces in state backends. I don't know if we use this anywhere, but it seems useful. If we want to get rid of this, I can put it behind a feature flag and check if the code still compiles.

Max-Meldrum: It is fine :) Just something to have in mind.

Max-Meldrum:
Is it ready to be merged or are there other changes you want to add?

mrobakowski:
Ready

PR #113, opened 2020-06-01
Don't use WAL in state backends

### LINKED ISSUE #112: Disable RocksDB WAL
The write-ahead log is enabled by default ( https://github.com/facebook/rocksdb/wiki/Write-Ahead-Log). I am 99% sure that we do not need it on as we operate within epochs. That is, we have the truth from the latest epoch checkpoint.

mrobakowski:
This disables the usage of write-ahead-log in rocks state backend. 

Unfortunately, rocksdb doesn't have any global config to disable WAL - you have to do that separately for every write-like operation. I hope I didn't miss any. Now that I think of it, maybe setting `wal_dir` to `/dev/null` would work. Probably not, because rocks would like to create and remove files there and that would fail.

Neither FASTER nor sled use WALs, so there are no changes there.

Closes #112

### DIFF
--- a/execution-plane/Cargo.lock
+++ b/execution-plane/Cargo.lock
@@ -103,7 +103,6 @@ dependencies = [
  "tui",
  "twox-hash",
  "uuid",
- "x86",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


Max-Meldrum:
Looks good :)

PR #114, opened 2020-06-11
NEXMark state backend metrics

mrobakowski:
This PR adds state backend metrics to the NEXMark experiments. A few new features had to be added to Kompact to make it all nice and tidy, so Kompact version is updated. I also managed to make the recovery tests not so flaky anymore.

### DIFF
--- a/execution-plane/arcon/src/test/recovery_tests.rs
+++ b/execution-plane/arcon/src/test/recovery_tests.rs
@@ -8,7 +8,8 @@ use std::{
     fs::File,
     io::{BufRead, BufReader},
     path::{Path, PathBuf},
-    sync::Mutex,
+    sync::{mpsc, Mutex},
+    time::Duration,
 };
 use tempfile::NamedTempFile;
 
@@ -107,10 +108,12 @@ fn run_pipeline<SB: state::Backend>(
         pool_info.clone(),
     ));
 
+    let (crash_tx, crash_rx) = mpsc::sync_channel(1);
     let crash_after = Mutex::new(crash_after);
     let map_fn = move |x: NormaliseElements| -> i64 {
         let mut crash_after = crash_after.lock().unwrap();
         if *crash_after == 0 {
+            crash_tx.send(()).unwrap();
             panic!("expected panic!")
         }
         *crash_after -= 1;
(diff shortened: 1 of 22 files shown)


execution-plane/arcon/src/test/recovery_tests.rs
```diff
@@ -270,7 +304,7 @@ fn test_sled_recovery_pipeline() {
 #[test]
 // flakey, sometimes passes, sometimes hangs, sometimes fails. But usually works as expected
 // when debugging :(
-#[ignore]
+// #[ignore]
```
Max-Meldrum, line 307: If the flaky part is fixed, then the comments can be removed.

Max-Meldrum:
Not sure I've fully understood the Dynamic Node part. What does it actually do? What is the difference?

mrobakowski:
> Not sure I've fully understood the Dynamic Node part. What does it actually do? What is the difference?

The proper, concrete node type is `Node<O, S, T>` where `O` is an operator, `S` is a state backend, and `T` is a timer backend. We want to change the state backend via a CLI flag, so we have to either: a) propagate all the generics up the call stack to right after we can dispatch the flag to a concrete type, or b) erase the concrete type of the node.

Option a) can cause a lot of code bloat and the code is more complicated. Option b) is viable only if we don't have to reason about the concrete type of the node. And we actually don't, because we only really care that it can be spawned/created on the kompact system.

You can see option a) in the `perf.rs` experiment, because I changed that one earlier. But it is way simpler and doesn't use the `NodeManager`, so it's not that big of a deal. But you can see that `main` is extremely short and basically just calls `exec`, which is generic, so the binary will have a separate copy of it for every state backend type (modulo optimization by the compiler and linker).

mrobakowski:
This also opens up the opportunity for one `NodeManager` to manage nodes with different state backends, which we probably want to do later. Previously `NodeManager` was parameterized with the type of the state backend of the managed nodes, but it didn't really need to know that, it was just to please the type system.

mrobakowski:
You could argue that it goes against what I did with the state backends inside of the node, which was the introduction of a new type parameter, or even two, on the `Node` struct, and holding a concrete backend type inline in `Node`. But there is one crucial difference here: inside the `Node`, state backend is referred to very often, it's the hot path. So removing an indirection and a virtual call there seems beneficial. Doing so in this case introduces no additional indirections **after** the node is created in the actor system. We just have some additional boxing going on before we spawn the nodes, but that happens only once at the beginning of the program.

PR #117, opened 2020-06-29
Channel Rework

### LINKED ISSUE #103: migrate to tell_serialised for remote channels
``tell_serialised`` demands a reference to a ComponentDefinition and not an ActorSource, so it will require some changes to the Channel Strategies.

For unit tests, we might have to do something like this.

```rust
src_comp.on_definition(|cd| {
  channel_strategy.add(ArconEvent::Element(element), &cd);
  // flush ...
});
```

Max-Meldrum:
This PR does the following:

1. Channel::Remote now uses the new `tell_serialised` function.
2. Remove flight serde bench. The difference between Reliable(Protobuf) and Unreliable(Abomonation) can already be seen in `benches/serde.rs`.

Closes #103


Max-Meldrum:
Will need some changes due to the changes in the Kompact code.

Max-Meldrum:
Merging this. I am currently not a big fan of the "mess" inside the ``node.rs`` file.  Mainly the abuse of RefCell and borrowing per operator call. I'll open another issue to tackle the Node -> OperatorContext  approach.

mrobakowski:
> I am currently not a big fan of the "mess" inside the `node.rs` file. Mainly the abuse of RefCell and borrowing per operator call.

I believe I am to blame for that 😞. https://github.com/cda-group/arcon/pull/110#issuecomment-631076245

PR #118, opened 2020-07-20
Put MapState operations that are slow on FASTER behind a feature flag.

mrobakowski:
Those operations will panic if the flag is off. Turning the flag on makes inserts very slow for big maps.


mrobakowski:
Ready to merge

PR #165, opened 2021-03-11
Add time helper methods

### LINKED ISSUE #162: Time helper methods
Add some time methods to make it easier when creating windows etc...

```rust
// src/stream/time/mod.rs
pub struct Time(u64);

impl Time {
 pub fn seconds(seconds: usize) -> Self { 
    // no need to convert...    
 }
 pub fn minutes(minutes: usize) -> Self {
    // convert to seconds
 }
 pub fn hours(hours: usize) -> Self {
    // convert to seconds
 }
 pub fn days(days: usize) -> Self {     
    // convert to seconds
 }
}
```

So rather than writing the seconds manually like this:

```rust
WindowAssigner::sliding(function, backend, 1000, 500, 0, false)
```

Use the time utility (just random values..):

```rust
WindowAssigner::sliding(function, backend, Time::hours(1), Time::hours(2), Time::minutes(10), false)
```

SoniaHorchidan:
Closes #162

### DIFF
--- a/guide/examples/src/bin/window.rs
+++ b/guide/examples/src/bin/window.rs
@@ -20,14 +20,27 @@ fn main() {
                 }
 
                 let function = IncrementalWindow::new(backend.clone(), &init, &aggregation);
-                WindowAssigner::sliding(function, backend, 1000, 500, 0, false)
+                WindowAssigner::sliding(
+                    function,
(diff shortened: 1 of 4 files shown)


Max-Meldrum:
Very nice!

PR #169, opened 2021-03-15
Add support for startup methods in operators

### LINKED ISSUE #159: Add support for startup methods in operators
Some operators may need to run startup code which has access to the `OperatorContext` before they begin to handle elements. For example, tumbling window operators might need to schedule a timer. Would it be possible to support an optional startup-method such as the following?

```rust
trait Operator {
    // ...
    fn startup(&mut self, mut ctx: OperatorContext<Self, impl Backend, impl ComponentDefinition>) -> OperatorResult<()> {
        Ok(())
    }
    // ...
}
```

If it is, then it could be a good first issue.

SoniaHorchidan:
Closes #159

### DIFF
--- a/src/stream/node/mod.rs
+++ b/src/stream/node/mod.rs
@@ -423,10 +423,7 @@ where
         unsafe {
             let operator = &mut (*self.operator.get());
             if let Err(_) = operator.on_start(make_context!(self)) {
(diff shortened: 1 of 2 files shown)


Max-Meldrum:
very nice 👍

Issue #164, opened 2021-03-11
Proc macro for deriving protobuf bloat code


segeljakt:
A copy of the `#[macros::proto]` macro in https://github.com/Max-Meldrum/arcon_taxi/tree/main/macros/src currently sits in the arc-script repo, but would you be interested in moving it to the arcon repo?

I think there is a lot of potential for using proc macros to automatically integrate Arcon with other tools. For example, it should be possible to derive a method for a data type which uploads a schema of that data type to Kibana, and also a function for uploading events of the schema.

Max-Meldrum:
Yes, I think that would be a nice addition! 

So I guess if it goes in  ``arcon_macros/``, one could use ``[arcon_macros::proto]``?

segeljakt:
sounds good 👍

PR #172, opened 2021-03-18
Add macro for deriving prost boilerplate

segeljakt:
This PR addresses #164. See `...` for comments.

### DIFF
--- a/arcon_macros/Cargo.toml
+++ b/arcon_macros/Cargo.toml
@@ -17,7 +17,7 @@ unsafe_flight = []
 arcon_arrow = []
 
 [dependencies]
-syn = "1.0"
+syn = { version = "1.0", features = ["full"] }
 quote = "1.0"
 proc-macro2 = "1.0"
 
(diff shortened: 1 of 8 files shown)


segeljakt:
The only caveat is that this PR pulls in `syn = { ... , features = ["full"] }`, is it OK or should we hide it behind a feature flag?

Max-Meldrum:
> The only caveat is that this PR pulls in `syn = { ... , features = ["full"] }`, is it OK or should we hide it behind a feature flag?

If it is needed for the PR, then I think its fine. If it slows down compilation too much we can look to remove it later on.

segeljakt:
Did I break something with Windows? 😬  @Max-Meldrum

Max-Meldrum:
> Did I break something with Windows? 😬 @Max-Meldrum

No, windows is broken 👍

Max-Meldrum (approved):
Looks good 👍  Closes #164

PR #188, opened 2021-05-12
Pull based Source Iterator

### LINKED ISSUE #182: Iterator interface for Sources

Max-Meldrum:
This PR changes the ``Source`` trait to be pulled based. The ``SourceNode`` component now poll a Source for records. The component now takes care of batching records and rescheduling itself rather than relying on the source implementation to "exit"... 

This is just a basic start of the ``Source`` trait, more methods (epoch, watermark) will most likely have to be added when we implement more advanced sources (e.g., Kafka).

Closes #182 

```rust
/// Enum containing Poll responses for an Arcon source
#[derive(Debug, Clone)]
pub enum Poll<A> {
    /// Makes the value `A` available
    Ready(A),
    /// Tells the runtime there is currently no records to process
    Pending,
    /// Indicates that the source is finished
    Done,
    /// An error occured while polling
    Error(String),
}


/// Defines an Arcon Source and the methods it must implement
pub trait Source: Send + 'static {
    type Item: ArconType;
    /// Poll Source for an Item
    fn poll_next(&mut self) -> Poll<Self::Item>;
    /// Set offset for the source
    ///
    /// May be used by replayable sources to set a certain offset..
    fn set_offset(&mut self, offset: usize);
}
```


segeljakt (approved):
LGTM

PR #200, opened 2021-06-18
Parallel kafka soucce

Max-Meldrum:
This PR adds local kafka parallel support.

Number of kafka partitions needs to be equal to or greater than the number of sources...

Using parallelism of 2 and 4 kafka partitions will lead to the following setup:

Source 0: partitions 0,1
Source 1: partitions 2,3

Still no handling of epochs or committing offsets.


Max-Meldrum:
Closes #73 Closes #144

PR #206, opened 2021-07-15
Remove nightly features

Max-Meldrum:
Now only depending on ``unboxed_closures`` in the arcon crate.


Max-Meldrum:
Turns out that with a few changes we can now compile on stable Rust. Moving forwards, we can still enable nightly features if there is a good reason for it.

Issue #217, opened 2021-07-20
User-defined Software Metrics


Max-Meldrum:
Part of #181

Allow users to define custom metrics in the Operator.

```rust
impl Operator for MyOperator {
    type IN = u64;
    type OUT = u64;
    type TimerState = ArconNever;
    type OperatorState = ();

    fn handle_element(
        &mut self,
        element: ArconElement<Self::IN>,
        mut ctx: OperatorContext<Self, impl Backend, impl ComponentDefinition>,
    ) -> ArconResult<()> {
       ctx.gauge("my_cool_gauge", 1000);
    }
  
    fn on_start(
        &mut self,
        mut ctx: OperatorContext<Self, impl Backend, impl ComponentDefinition>,
    ) -> ArconResult<()> {
       ctx.register_gauge("my_cool_gauge");
    }
    ignore_timeout!();
    ignore_persist!();
}
```

Issue #216, opened 2021-07-19
Hardware Metrics


Max-Meldrum:
Part of #181 

Allow users to enable and customise hardware counters around an Operator if a  ``hardware_counter`` feature flag is on.

*  CACHE MISSES
*  BRANCH MISSES
*  INSTRUCTIONS
*  CPU CYCLES
*  ....

```rust
let mut perf_events = PerfEvents::new();
perf_events.add(HardwareCounter::CPU_CYCLES);
perf_events.add(HardwareCounter::CACHE_MISSES);

let builder = OperatorBuilder {
   constructor: Arc::new(|_| Map::new(|x| x + 10)),
   conf: OperatorConf {
       perf_events,
       ..Default::default()
   },
};
````

Issue #215, opened 2021-07-19
Software Metrics


Max-Meldrum:
Part of #181

* Node
    *  Events Throughput
    *  Watermark Gauge
    *  Epoch Gauge
    *  Inbound/Outbound Channels Gauge
* NodeManager
    *  Checkpoint Execution Time
*  Source
    * Records Throughput
    * Poll Error Counter
    * Watermark Gauge
* Index/Backend
    * ?

Issue #181, opened 2021-04-09
Add metrics integration


Max-Meldrum:
*  Record runtime metrics
*  Allow users to enable hardware counters
*  Allow users to register custom software metrics in an Operator
*  Metrics Visualisation

Max-Meldrum:
https://github.com/jimblandy/perf-event

PR #225, opened 2021-07-23
Metrics Integration

Sanskar95:
This PR integrate two types of metrics: Hardware counters Software metrics enabled using following feature flags respectively.

 **hardware_counters**
     - Enables counters like cache misses, branch misses etc.
     - It is to be noted that this feature is only compatible with linux OS as it uses perf_event_open() under the hood
     - One has to provide CAP_SYS_ADMIN capability to use it for eg:  setcap cap_sys_admin+ep target/debug/collection , this takes the built file as an argument.
     - Not executing the above command will result into "Operation not permitted" error assuming the feature flag is enabled.
       
     
**metrics**
    - Enables software metrics for node like epoch_counter, water_mark_counter etc.
    - It also enables source node realted metrics like inbound_throughput, error_counter etc.
    - One can also register custom counters using the operator_context if the feature flag is on.


Max-Meldrum:
Nice work! 

Below are some high-level comments, I'll also add some smaller review comments later.


# Feature flags and default recorder

We can remove the arcon_metrics feature as we are not using metrics-util + that I think prometheus exporter should be its own feature flag under "prometheus_exporter".
That is, if only metrics flag is enabled then we should default to a LogRecorder.

prometheus_exporter = ["metrics-exporter-prometheus", "metrics"]
hardware_counters = ["perf-event", "metrics"]

```rust
// in src/metrics/
pub struct LogRecorder {
  logger: ArconLogger,
}

impl Recorder for LogRecorder {
   fn increment_counter(&self, key: &Key, value: u64) {
      info!(self.logger, "Key {}, Value {}", key, value);
   }
   ... so on
}
```

```rust
// Application::new()
#[cfg(all(feature = "prometheus_exporter", feature = "metrics"))]
// Prometheus recorder
#[cfg(feature = "metrics")]
// LogRecorder
```


# Runtime_Metrics

For this part, there is no need to have the MetricValue trait and individual implementations of Counters. We can
use the ``metrics`` macros directly.

```rust
// Node Setup
register_counter!("node_name_id_epoch_counter");
register_counter!("node_name_id_watermark_counter");
// Node Execution
increment_counter!("node_name_id_epoch_counter");
increment_counter!("node_name_id_watermark_counter");
```

We can still keep a NodeMetrics struct for other stuff that is needed. Like the Meter.

```rust
pub struct NodeMetrics {
  inbound_throughput: Meter,
}
// Node
self.metrics.inbound_throughput.mark_n(message.total_events());
gauge!("node_name_id_inbound_throughput", self.metrics.inbound_throughput.get_mean_rate());
```


# Histogram

1. I realised we should probably use histogram for perf-events rather than gauge. 
2. We might as well add histogram to the OperatorContext while we are at it.

Cargo.toml
```diff
@@ -32,7 +32,9 @@ unsafe_flight = ["abomonation", "abomonation_derive", "arcon_macros/unsafe_fligh
 kafka = ["rdkafka", "futures", "serde_json", "serde"]
 thread_pinning = ["kompact/thread_pinning"]
 socket = ["tokio-util", "futures", "serde_json", "serde"]
-metrics = []
+arcon_metrics=["metrics", "metrics-exporter-prometheus", "metrics-util"]
+hardware_counters=["perf-event", "arcon_metrics"]
```
Max-Meldrum, line 36: Not sure if there is a formatter for Cargo.toml, but could u fix the spacing? e.g., hardware_counters = ["..."] and not hardware_counters=[".."]

Cargo.toml
```diff
@@ -81,6 +83,15 @@ rayon = { version = "1.3.0", optional = true }
 abomonation = { version = "0.7.3", optional = true }
 abomonation_derive = { version = "0.5.0", optional = true }
 
+
+metrics = {version="0.16.0", optional=true}
```
Max-Meldrum, line 87: The spacing here too { version = "...", } and not {version="..."}.

src/application/mod.rs
```diff
@@ -102,6 +106,14 @@ impl Default for Application {
 impl Application {
     /// Creates a new Application using the given ApplicationConf
     fn new(conf: ApplicationConf) -> Self {
+        #[cfg(feature = "metrics")]
```
Max-Meldrum, line 109: As mentioned in the comment. Default to a LogRecorder if only the metrics flag is enabled. And have prometheus feature under ``prometheus_exporter``.

src/dataflow/conf.rs
```diff
@@ -83,6 +88,7 @@ pub struct SourceConf<S: ArconType> {
     pub extractor: Option<TimestampExtractor<S>>,
     pub time: ArconTime,
     pub batch_size: usize,
+    pub source_name: String,
```
Max-Meldrum, line 91:
```rust
pub name: String,  and pub fn set_name(....).
```

src/dataflow/constructor.rs
```diff
@@ -296,6 +296,9 @@ pub(crate) fn node_manager_constructor<OP: Operator + 'static, B: Backend>(
                     NodeState::new(node_id, in_channels.clone(), backend.clone()),
                     backend.clone(),
                     app.arcon_logger.clone(),
+                    #[cfg(all(feature = "hardware_counters", target_os = "linux"))]
```
Max-Meldrum, line 299: #[cfg(all(feature = "hardware_counters", target_os = "linux", not(test)))]

src/lib.rs
```diff
@@ -26,6 +26,16 @@
 //!     - Enables RocksDB to be used as a Backend
 //! - `arcon_serde`
 //!     - Adds serde support for Arcon Types
+//! - `hardware_counters`
+//!     - Enables counters like cache misses, branch misses etc.
+//!     - It is to be noted that this feature is only compatible with linux OS as it uses perf_event_open() under the hood
+//!     - One has to provide CAP_SYS_ADMIN capability to use it for eg:  setcap cap_sys_admin+ep target/debug/collection , this takes the built file as an argument.
+//!     - Not executing the above command will result into "Operation not permitted" error assuming the feature flag is enabled.
+//! - `metrics'
+//!     - Enables software metrics for node like epoch_counter, water_mark_counter etc.
```
Max-Meldrum, line 35: The hardware_counters section looks good. But could be good to add the metrics one above it and then have ``hardware_counters`` and ``prometheus_exporter`` below.

Also for metrics, this text is enough:
- Records internal runtime metrics and allows users to register custom metrics from an Operator
- If no exporter (e.g., prometheus_exporter) is enabled, the metrics will be logged by the runtime.

src/metrics/ewma.rs
```diff
@@ -15,7 +16,7 @@ const M5_ALPHA: f64 = 1.0 - (-INTERVAL as f64 / SECONDS_PER_MINUTE / FIVE_MINUTE
 const M15_ALPHA: f64 = 1.0 - (-INTERVAL as f64 / SECONDS_PER_MINUTE / FIFTEEN_MINUTES as f64);
 
 /// Exponentially Weighted Moving Average
-#[derive(Clone, Debug)]
+#[derive(Deserialize, Clone, Debug, Default)]
```
Max-Meldrum, line 19: Why do ``EWMA`` and ``Meter`` need Deserialize?

src/metrics/mod.rs
```diff
@@ -3,13 +3,10 @@
...
-/// Histogram implementation
-pub mod histogram;
-/// Meter implementation
-pub mod meter;
+mod ewma;
+mod meter;
+#[cfg(all(feature = "hardware_counters", target_os = "linux"))]
+pub mod perf_event;
+
+#[cfg(feature = "metrics")]
```
Max-Meldrum, line 11: the ``feature = metrics`` part won't be needed here as the metrics module should be covered under the feature in src/lib.rs

src/metrics/perf_event.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub fn new() -> PerfEvents {
+        PerfEvents { counters: vec![] }
+    }
+
+    pub fn add(&mut self, hardware_metric_kind: HardwareCounter) {
+        self.counters.push(hardware_metric_kind);
+    }
+}
+
+pub struct HardwareMetricGroup {
```
Max-Meldrum, line 70: As mentioned in the comment.  Move towards using histogram instead.

src/metrics/runtime_metrics.rs
```diff
@@ -0,0 +1,148 @@
+use crate::metrics::meter::Meter;
```
Max-Meldrum, line 1: As mentioned in the larger comment. Get rid of MetricValue and individual counters and IncomingMessageRate struct and use ``metrics`` macros directly where possible or use ``Meter`` directly instead of wrapped in ``IncomingMessageRate``.

src/stream/node/mod.rs
```diff
@@ -6,6 +6,13 @@ pub mod debug;
 
 pub mod source;
 
+#[cfg(feature = "metrics")]
+use metrics::{counter, gauge};
+
+#[cfg(all(feature = "hardware_counters", target_os = "linux"))]
```
Max-Meldrum, line 12: #[cfg(all(feature = "hardware_counters", target_os = "linux", not(test)))]

src/stream/node/mod.rs
```diff
@@ -28,54 +35,15 @@ use fxhash::*;
...
-    pub inbound_channels: Gauge,
-}
-
-#[cfg(feature = "metrics")]
-impl Default for NodeMetrics {
-    fn default() -> Self {
-        Self::new()
-    }
-}
+#[cfg(all(feature = "hardware_counters", target_os = "linux"))]
```
Max-Meldrum, line 41: Same here

src/stream/node/mod.rs
```diff
@@ -101,6 +69,7 @@ pub struct NodeState<OP: Operator + 'static, B: Backend> {
     /// Identifier for the Node
     #[ephemeral]
     id: NodeID,
+    // perf_metrics_group: Group,
```
Max-Meldrum, line 72: remove commented code

src/stream/node/mod.rs
```diff
@@ -161,14 +131,19 @@ where
...
-    /// Metrics collected by the Node
-    metrics: NodeMetrics,
+
     /// Internal Node State
     node_state: NodeState<OP, B>,
     /// Event time scheduler
     timer: UnsafeCell<ArconTimer<u64, OP::TimerState, B>>,
     logger: ArconLogger,
+
+    #[cfg(all(feature = "hardware_counters", target_os = "linux"))]
```
Max-Meldrum, line 141: same here

src/stream/node/mod.rs
```diff
@@ -218,28 +237,39 @@ where
...
-        match message {
-            MessageContainer::Raw(r) => self.handle_events(r.sender, r.events),
-            MessageContainer::Local(l) => self.handle_events(l.sender, l.events),
+        #[cfg(all(feature = "hardware_counters", target_os = "linux"))]
+        #[cfg(not(test))]
+        {
+            let counts = self.hardware_metric_group.group.read()?;
+            let counter_iterator = self.hardware_metric_group.counters.iter();
+            for (metric_name, counter) in counter_iterator {
+                gauge!(
```
Max-Meldrum, line 257: The histogram change + don't we need to use .reset() afterward?.

```rust
self.hardware_metric_group.group.reset()?;
```

src/stream/operator/mod.rs
```diff
@@ -124,12 +130,16 @@ where
         timer: &'b mut Timer<u64, OP::TimerState, B>,
         channel_strategy: &'c mut ChannelStrategy<OP::OUT>,
         logger: &'d ArconLogger,
+        _name: &'a str,
```
Max-Meldrum, line 133: Add feature flag above name parameter? in the make_context!() macro you can also add the flag above the borrow to &sel.descriptor etc..

src/metrics/log_recorder.rs
```diff
@@ -0,0 +1,41 @@
+use crate::application::conf::logger::ArconLogger;
+use metrics::{GaugeValue, Key, Recorder, Unit};
+
+pub struct LogRecorder {
+    pub(crate) logger: ArconLogger,
+}
+
+impl Recorder for LogRecorder {
+    fn register_counter(&self, key: &Key, unit: Option<Unit>, description: Option<&'static str>) {
+        println!(
```
Max-Meldrum, line 10: the logger is not being used 😄  Just replace the println!  with ``info!(self.logger,  ....);``

src/application/mod.rs
```diff
@@ -105,17 +107,27 @@ impl Default for Application {
 
 impl Application {
     /// Creates a new Application using the given ApplicationConf
+
     fn new(conf: ApplicationConf) -> Self {
-        #[cfg(feature = "metrics")]
+        #[cfg(all(feature = "prometheus_exporter", feature = "metrics", not(test)))]
         {
             let _builder = PrometheusBuilder::new();
```
Max-Meldrum, line 114: if we are not storing the metrics recorder/builder ``PrometheusBuilder`` then just install directly without variables.

```rust
PrometheusBuilder::new()
  .install()
  .expect("failed to install Prometheus recorder");
```

src/stream/node/mod.rs
```diff
@@ -237,29 +232,41 @@ where
             self.node_state.message_buffer().append(message.raw())?;
             return Ok(());
         }
+
+        #[cfg(feature = "metrics")]
+        self.node_metrics.start_timer();
```
Max-Meldrum, line 237: As this measurement is just a local concern, there is no need to store it in NodeMetrics. Use ``std::time::Instant`` like below.  The only thing is turning the elapsed time into an f64 as the histogram requires it I think?  there is a ``as_secs_f64`` function on Duration, but we might want it in something like nanoseconds/milliseconds.

```rust
#[cfg(feature = "metrics")]
let start = Instant::now();

match message {
   MessageContainer::Raw(r) => self.handle_events(r.sender, r.events),
   MessageContainer::Local(l) => self.handle_events(l.sender, l.events),
}

#[cfg(feature = "metrics")]
{
  let elapsed = start.elapsed(); //  needs to be converted to f64 for histogram?
  histogram!("....", elapsed);
}
```

Max-Meldrum:
Looks good!

Closes #181, closes #215, closes #216, closes #217

Issue #231, opened 2021-07-29
metrics refactoring


Max-Meldrum:
# Move metrics registration out of NodeMetrics/SourceMetrics

This should happen in the construction of SourceNode/Node (in new(...)).

The structs should only hold data structures used for metrics (e.g., Meter).

# Change how we use ``metrics`` macro

Rather than combining strings each time, we can use the macro key/labels.

```rust
// old
register_histogram!(format!("{}_{}", node_name, "batch_execution_time"));
// new
register_histogram!(
  "batch_execution_time",
   Unit::Microseconds,
   "execution time per events batch",
   "node" => self.descriptor.clone(),
);

// old
histogram!(
  format!("{}_{}", &self.descriptor, "batch_execution_time"),
  elapsed.as_micros() as f64
);
// new
histogram!("batch_execution_time",  elapsed.as_micros() as f64, "node" => self.descriptor.clone());


// new counter
increment_counter!("watermark_counter", "node" => self.descriptor.clone());

```

# Source throughput per batch

Currently, we are marking and reporting the Meter metric per Polled record. Change the ``process`` function to return ArconResult<usize> where usize is the counter variable indicating how many records were polled.  

```rust
match self.process() {
#[cfg(not(feature = "metrics")]
Ok(_) => (),
#[cfg(feature = "metrics")]
Ok(polled_records) => {
   self.metrics.incoming_message_rate.mark_n(polled_records);
   gauge!("incoming_message_rate",  self.metrics.incoming_message_rate.one_min_rate(), "source" => self.descriptor().clone());
}
Err(error) => {
 // fatal error, must shutdown..
 // TODO: coordinate shutdown of the application..
 error!(self.logger, "{}", error);
}
```

## Update HardwareMetricGroup

function names contain term gauge.

I think we can remove both of the functions if we are moving the registration to  Node::new() and using the new macro approach.

PR #234, opened 2021-07-29
Used labels and service while calling metrics-rs macros

Sanskar95:


src/metrics/perf_event.rs
```diff
@@ -80,12 +80,8 @@ impl HardwareMetricGroup {
     ) -> std::io::Result<()> {
         let iterator = perf_events.counters.iter();
         for value in iterator {
-            register_histogram!(self.get_field_gauge_name(&node_name, &value.to_string()));
+            register_histogram!(value.to_string(),"node" => String::from (node_name.clone()));
         }
         self.group.enable()
```
Max-Meldrum, line 85: ah didn't catch this one before. Why is the group enabled here?

But could we remove the whole function? And move the registration out of the struct and handle it in the Node.

```rust
// in Node::new()
for counter in hardware_metric_group.counters() {
  register_histogram!(counter, "node" => self.descriptor.clone());
}
```

src/stream/node/source.rs
```diff
@@ -129,31 +129,37 @@ where
...
-                    return Ok(());
+                    return Ok(counter);
                 }
                 Err(error) => {
                     #[cfg(feature = "metrics")]
-                    increment_counter!(format!("{}_{}", &self.descriptor, "error_counter"),);
-                    return self.handle_source_error(error);
+                    increment_counter!("error_counter", "source" => self.descriptor.clone());
+
+                    return self.handle_source_error(error, error_counter);
```
Max-Meldrum, line 143: As discussed. no need to have counter variable within the ``handle_source_error`` function. 
Add something like this where we simply bump the counter if the error is acceptable, but otherwise return Err if it is seen as critical.
```rust
match self.handle_source_error(error) {
    Ok(_) => {
        counter += 1; // Acceptable Error, just increase counter
    }
    Err(err) {
        return Err(err); // Critical error 
    }
}
```

Sanskar95:
Fixes https://github.com/cda-group/arcon/issues/231

src/stream/node/mod.rs
```diff
@@ -196,9 +198,15 @@ where
...
-                .register_hardware_metric_gauges(descriptor.clone(), perf_events)
-                .ok();
+            // hardware_metric_group
+            //     .register_hardware_metric_gauges(descriptor.clone(), perf_events)
+            //     .ok();
+            let iterator = perf_events.counters.iter();
+            for value in iterator {
+                register_histogram!(value.to_string(),"node" => descriptor.clone());
+            }
+            hardware_metric_group.group.enable().ok();
```
Max-Meldrum, line 206: should we be enabling the group here in the constructor?

Well to check that there are no errors yes, but it will start counting.  Its probably fine to have it there, but then we should probably move the group.reset()?; before rather than after executing handle_events(...).

```rust
#[cfg(feature = "hardware_counters", ... )]
self.hardware_metric_group.reset()?;
 #[cfg(feature = "metrics")]
let start_time = Instant::now();
```

src/stream/node/mod.rs
```diff
@@ -249,13 +218,52 @@ where
             return Ok(());
         }
 
+        #[cfg(all(feature = "hardware_counters", target_os = "linux", not(test)))]
```
Max-Meldrum, line 221: registration should happen in Node::new().

I think we can remove the HardwareMetricGroup now that we operate locally. Also, a lot of code can be removed.

We don't need to reset the group.

for iterators, no need to store the iterator in a variable.
```rust
for counter in self.perf_events.counters.iter() {
}
// instead of
let iterator = self.perf_events.counters.iter();
for counter in iterator {

}
```

The code below has not been verified to compile but should be possible to write it like that.





```rust
// Node constructor
#[cfg(all(feature = "hardware_counters", target_os = "linux", not(test)))]
for counter in self.perf_events.counters.iter() {
    register_histogram!(counter.to_string(),"node" => self.descriptor.clone());
}

// handle message

#[cfg(all(feature = "hardware_counters", target_os = "linux", not(test)))]
let mut (group, counters) = {
    let group = Group::new()?;
    let counters = Vec::with_capacity(self.perf_events.counters.len());
    for hardware_counter in self.perf_events.counters.iter() {
        let counter = Builder::new()
                        .group(&mut hardware_metric_group.group)
                        .kind(val.get_hardware_kind())
                        .build()?;
                        
        counters.push((hardware_counter.to_string(), counter));
    }
    (group, counters)
};

#[cfg(all(feature = "hardware_counters", target_os = "linux", not(test)))]
group.enable()?;

// do work

#[cfg(all(feature = "hardware_counters", target_os = "linux", not(test)))]
group.disable()?;

#[cfg(all(feature = "hardware_counters", target_os = "linux", not(test)))]
let counts = group.read()?;

#[cfg(all(feature = "hardware_counters", target_os = "linux", not(test)))]
for (metric_name, counter) in counters.iter() {
  histogram!(String::from(metric_name), counts[counter] as f64, "node" => self.descriptor.clone());
}
```

Issue #239, opened 2021-08-09
Allocator metrics


Max-Meldrum:
Track information of  ``arcon_allocator``.

* total_bytes (gauge) 
* bytes_remaining (gauge)
* alloc_counter (counter)

Issue #238, opened 2021-08-09
Manager metrics


Max-Meldrum:
*  SourceManager
   *  sources (gauge)
*  NodeManager
   *   checkpoint_execution_time_ms (histogram) (milliseconds)
   *   nodes (gauge)

PR #240, opened 2021-08-09
 Added Allocator and manager metrics

Sanskar95:


Sanskar95:
Fixes
https://github.com/cda-group/arcon/issues/239
https://github.com/cda-group/arcon/issues/238

PR #241, opened 2021-08-09
added manager and allocator metrics

### LINKED ISSUE #238: Manager metrics

Sanskar95:
Fixes
#238 
#239

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,11 +34,10 @@ thread_pinning = ["kompact/thread_pinning"]
 socket = ["tokio-util", "futures", "serde_json", "serde"]
 hardware_counters = ["perf-event", "metrics"]
 prometheus_exporter = ["metrics-exporter-prometheus", "metrics"]
-
+allocator_metrics = ["arcon_allocator/metrics"]
 
 # state features
 rocksdb = ["arcon_state/rocks"]
-arcon_allocator_counters = ["arcon_allocator/allocator_metrics"]
 
 [dependencies]
 arcon_macros = { version = "0.1", path = "arcon_macros" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ prometheus_exporter = ["metrics-exporter-prometheus", "metrics"]
 
 # state features
 rocksdb = ["arcon_state/rocks"]
+arcon_allocator_counters = ["arcon_allocator/allocator_metrics"]
 
 [dependencies]
 arcon_macros = { version = "0.1", path = "arcon_macros" }
--- a/arcon_allocator/Cargo.toml
+++ b/arcon_allocator/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Max Meldrum <mmeldrum@kth.se>"]
 edition = "2018"
 
 [features]
-allocator_metrics = ["metrics-exporter-prometheus", "metrics"]
+allocator_metrics = ["metrics"]
 
 
 [dependencies]
@@ -13,4 +13,3 @@ fxhash = "0.2.1"
 snafu = "0.6"
 
 metrics = { version = "0.16.0", optional = true }
-metrics-exporter-prometheus = { version = "0.5.0", optional = true }
--- a/arcon_allocator/Cargo.toml
+++ b/arcon_allocator/Cargo.toml
@@ -4,10 +4,6 @@ version = "0.1.0"
 authors = ["Max Meldrum <mmeldrum@kth.se>"]
 edition = "2018"
 
-[features]
-allocator_metrics = ["metrics"]
-
-
 [dependencies]
 fxhash = "0.2.1"
 snafu = "0.6"
--- a/arcon_allocator/Cargo.toml
+++ b/arcon_allocator/Cargo.toml
@@ -4,6 +4,13 @@ version = "0.1.0"
 authors = ["Max Meldrum <mmeldrum@kth.se>"]
 edition = "2018"
 
+[features]
+allocator_metrics = ["metrics-exporter-prometheus", "metrics"]
+
+
 [dependencies]
 fxhash = "0.2.1"
 snafu = "0.6"
+
+metrics = { version = "0.16.0", optional = true }
+metrics-exporter-prometheus = { version = "0.5.0", optional = true }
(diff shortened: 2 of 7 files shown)


arcon_allocator/Cargo.toml
```diff
@@ -4,6 +4,13 @@ version = "0.1.0"
 authors = ["Max Meldrum <mmeldrum@kth.se>"]
 edition = "2018"
 
+[features]
+allocator_metrics = ["metrics-exporter-prometheus", "metrics"]
```
Max-Meldrum, line 8: We don't need the prom exporter in the ``arcon_allocator`` crate so having metrics as optional dependency only is fine.  Which can be activated through ``arcon_allocator/metrics``

Sanskar95: yeah good point!

arcon_allocator/src/lib.rs
```diff
@@ -56,6 +59,13 @@ pub struct Allocator {
 impl Allocator {
     /// Creates a new Allocator with the given memory limit size
     pub fn new(limit: usize) -> Allocator {
+        #[cfg(feature = "allocator_metrics")]
+        {
+            register_gauge!("total_bytes", "arcon_allocator" => "arcon_allocator");
```
Max-Meldrum, line 64: Do we need the "arcon_allocator" => "arcon_allocator" part?

```rust
register_gauge!("bytes_remaining",  "allocator");
// or 
register_gauge!("allocator_bytes_remaining");


```

Sanskar95: I added that to maintain consistency with counters registration as per the node metrics

arcon_allocator/src/lib.rs
```diff
@@ -110,6 +120,13 @@ impl Allocator {
         self.allocations
             .insert((self.alloc_epoch, id), (mem, layout));
 
+        #[cfg(feature = "allocator_metrics")]
+        {
+            increment_counter!("alloc_counter", "arcon_allocator" => "arcon_allocator");
+            gauge!("bytes_remaining",self.bytes_remaining() as f64 ,"arcon_allocator" => "arcon_allocator");
+            gauge!("total_bytes",self.limit as f64 ,"arcon_allocator" => "arcon_allocator");
```
Max-Meldrum, line 127: For now the total_bytes will always be the same as the initial limit in the ``Allocator::new(..)`` so let's move the gauge call up there.

src/manager/source.rs
```diff
@@ -77,6 +85,11 @@ impl<B: Backend> SourceManager<B> {
     }
 
     pub(crate) fn add_source(&mut self, source: Arc<dyn AbstractComponent<Message = SourceEvent>>) {
+        #[cfg(feature = "metrics")]
```
Max-Meldrum, line 88: We add the gauge call to the handling of Start message.

```rust
if SourceEvent::Start == msg {
 #[cfg(feature = "metrics")]
  gauge!("sources", self.nodes.len() as f64 "source_manager" => self.state_id.clone() );
}
```

src/manager/node.rs
```diff
@@ -254,6 +264,8 @@ where
     }
 
     fn handle_node_event(&mut self, event: NodeManagerEvent) -> ArconResult<()> {
+        gauge!("nodes", self.nodes.len() as f64 ,"node_manager" => self.state_id.clone());
```
Max-Meldrum, line 267: Probably don't want to update the gauge every ``handle_node_event`` call.  As the nodes are currently static, calling the gauge!(..) at on_start would be suitable. However, I think the manager currently starts before actually adding the nodes into ``self.nodes``.

For it to work at ``on_start``, the following change has to happen:

```rust
// Start NodeManager
  app.ctrl_system()  // Line 270 src/dataflow/constructor.rs
     .start_notify(&manager_comp)
     .wait_timeout(std::time::Duration::from_millis(2000))
     .expect("Failed to start NodeManager");
// has to be moved down to after line 320.
```

arcon_allocator/Cargo.toml
```diff
@@ -5,12 +5,11 @@ authors = ["Max Meldrum <mmeldrum@kth.se>"]
 edition = "2018"
 
 [features]
-allocator_metrics = ["metrics-exporter-prometheus", "metrics"]
+allocator_metrics = ["metrics"]
```
Max-Meldrum, line 8: The feature can be removed and we can use just "metrics".

src/manager/source.rs
```diff
@@ -134,6 +127,9 @@ impl<B: Backend> Actor for SourceManager<B> {
         // If we received a start message, start the periodic timer
         // that instructs sources to send off watermarks.
         if SourceEvent::Start == msg {
+            #[cfg(feature = "metrics")]
+            gauge!("sources", 1.0,"source_manager" => self.state_id.clone());
```
Max-Meldrum, line 131: It seems like a typo here.  ``1.0`` instead of ``self.sources.len() as f64``

src/dataflow/constructor.rs
```diff
@@ -273,6 +274,9 @@ pub(crate) fn node_manager_constructor<OP: Operator + 'static, B: Backend>(
                 .wait_timeout(std::time::Duration::from_millis(2000))
                 .expect("Failed to start NodeManager");
 
+            #[cfg(feature = "metrics")]
```
Max-Meldrum, line 277: Well this works too, but I meant moving the gauge! call to the NodeManager's ``on_start`` method. And described what code had to be moved in the constructor part for it to work properly.

Would preferably like to keep out metrics calls outside of dataflow construction.

Cargo.toml
```diff
@@ -38,6 +38,7 @@ prometheus_exporter = ["metrics-exporter-prometheus", "metrics"]
 
 # state features
 rocksdb = ["arcon_state/rocks"]
+arcon_allocator_counters = ["arcon_allocator/allocator_metrics"]
```
Max-Meldrum, line 41: arcon_allocator_counters -> allocator_metrics = ["arcon_allocator/metrics"]

And let's move it above the state features comment in order to not cause confusion.

Max-Meldrum:
Looks good. Nice work!

Issue #244, opened 2021-08-13
Remove unecessary gauge call from node manager


Sanskar95:

Max-Meldrum:
Closing issue. @Sanskar95  Using Closes #issue takes care of automatically closing the issues once the PR is merged.

PR #245, opened 2021-08-13
removed redundant gauge call from node manager

Sanskar95:


Sanskar95:
Fixes #244

Max-Meldrum:
Great! There is a warning for unused import  ``use std::time::Insant`` line 27 in src/manager/node.rs when running with default features.  The import should be under ``#[cfg(feature = "metrics")]``

And while you are at it, could you add in ``run_checks.sh`` another line that runs clippy without ``all-features``.

```bash
cargo clippy --workspace --all-targets  -- -D warnings
```
We probably need this check at the CI level as well, but useful for catching potential warnings locally.

Issue #242, opened 2021-08-11
State Metrics


Max-Meldrum:
Add ``metrics`` as an optional dep to ``arcon_state``. 

We most likely have to give each ``Backend`` the ID of the operator when constructing it.

Then we have to figure out where and how often metrics should be reported.  ``Handle`` might be a suitable location.

PR #249, opened 2021-08-17
arcon_state metrics

Sanskar95:


arcon_state/src/backend/sled/vec_ops.rs
```diff
@@ -33,7 +35,11 @@ impl VecOps for Sled {
             <usize as FixedBytes>::SIZE + protobuf::size_hint(&value).unwrap_or(0),
         );
         #[cfg(feature = "metrics")]
-        counter!(format!("{}_bytes_written", handle.get_name()), serialized.len() as u64, "backend" => self.name.clone());
+        record_bytes_written(
```
Max-Meldrum, line 38: move the record call under the 2 serialize_into calls.

arcon_state/Cargo.toml
```diff
@@ -12,7 +12,7 @@ State management features for Arcon
 """
 
 [features]
-default = ["sled", "sled_checkpoints"]
+default = ["sled", "sled_checkpoints", "metrics"]
```
Max-Meldrum, line 15: metrics should not be a default feature for ``arcon_state``

We need to introduce a new ``state_metrics`` feature in the core Cargo.toml

arcon_state/src/backend/mod.rs
```diff
@@ -124,16 +124,18 @@ pub trait Backend:
...
-                Self::restore(&state_path, &latest_checkpoint_path)
+                Self::restore(&state_path, &latest_checkpoint_path, id)
             }
-            None => Self::create(&state_path),
+            None => Self::create(&state_path, id),
         }
     }
 
-    fn create(live_path: &Path) -> Result<Self>
+    fn return_name(&self) -> String;
```
Max-Meldrum, line 133: Let's use string literals instead and avoid cloning everywhere.

And no need to have return in the function name.

```rust
fn name(&self) -> &str;
```

arcon_state/src/backend/rocks/mod.rs
```diff
@@ -26,6 +27,7 @@ pub struct Rocks {
     inner: UnsafeCell<DB>,
     path: PathBuf,
     restored: bool,
+    name: String,
```
Max-Meldrum, line 30:
```rust
name: &'static str
```

arcon_state/src/backend/sled/mod.rs
```diff
@@ -20,6 +20,7 @@ use std::{
 pub struct Sled {
     db: Db,
     restored: bool,
+    name: String,
```
Max-Meldrum, line 23:
```rust
name: &'static str
```

arcon_state/src/backend/metrics_utils.rs
```diff
@@ -0,0 +1,14 @@
+#[cfg(feature = "metrics")]
```
Max-Meldrum, line 1: You can remove all the cfg feature lines here in this file. 
You just need to have it covered under the flag when declaring the module.

```rust
// src/backend/mod.rs
#[cfg(feature = "metrics")]
pub mod metrics_utils;
```

src/manager/node.rs
```diff
@@ -305,7 +314,16 @@ where
 
                             if OP::OperatorState::has_tables() {
                                 if let Some(snapshot) = &self.latest_snapshot {
+                                    #[cfg(feature = "metrics")]
```
Max-Meldrum, line 317: This part of the code will most likely be removed in the coming future. So best to remove everything regarding the last_checkpoint_restore_timestamp metric

arcon_state/src/backend/handles.rs
```diff
@@ -235,6 +242,12 @@ impl<S: StateType, IK: Metakey, N: Metakey> Handle<S, IK, N> {
             panic!("State handles should be registered before activation!")
         }
 
+        #[cfg(feature = "metrics")]
```
Max-Meldrum, line 245: As the bytes metrics are for a whole backend and not individual handles. Let's move the registrations to the creation of the Backend

Max-Meldrum: ah nevermind, it is for individual handles. My bad

Max-Meldrum: Ignore the comment 😄

Cargo.toml
```diff
@@ -35,6 +35,7 @@ socket = ["tokio-util", "futures", "serde_json", "serde"]
 hardware_counters = ["perf-event", "metrics"]
 prometheus_exporter = ["metrics-exporter-prometheus", "metrics"]
 allocator_metrics = ["arcon_allocator/metrics"]
+state_metrics=["arcon_state/metrics"]
```
Max-Meldrum, line 38:
```suggestion
state_metrics = ["arcon_state/metrics"]
```

arcon_state/src/backend/handles.rs
```diff
@@ -225,6 +228,10 @@ impl<S: StateType, IK: Metakey, N: Metakey> Handle<S, IK, N> {
     pub fn metakey_size(&self) -> usize {
         IK::SIZE + N::SIZE
     }
+
+    pub fn get_name(&self) -> String {
```
Max-Meldrum, line 232:
```suggestion
    pub fn get_name(&self) -> &str {
```

arcon_state/src/backend/metrics_utils.rs
```diff
@@ -0,0 +1,13 @@
+use metrics::counter;
```
Max-Meldrum, line 1: metrics_utils file does not need cfg feature calls as mentioned by the previous comment.

arcon_state/src/backend/rocks/mod.rs
```diff
@@ -139,7 +141,11 @@ where
 }
 
 impl Backend for Rocks {
-    fn create(path: &Path) -> Result<Self>
+    fn return_name(&self) -> &'static str {
```
Max-Meldrum, line 144:
```suggestion
    fn name(&self) -> &str {
```

benches/appender.rs
```diff
@@ -58,7 +58,13 @@ fn mean(numbers: &[u64]) -> f32 {
 fn appender_mean_eager(backend: BackendType, window_size: usize, b: &mut Bencher) {
     let dir = tempdir().unwrap();
     with_backend_type!(backend, |B| {
-        let backend = Arc::new(B::create(dir.as_ref()).unwrap());
+        let backend = Arc::new(
+            B::create(
+                dir.as_ref(),
+                Box::leak(backend.to_string().into_boxed_str()),
```
Max-Meldrum, line 64:
```suggestion
                "appender_backend"
```

This is just benchmarks, so the name of the backend can just be anything. For example, name it the same as the type of benchmark.

src/dataflow/stream.rs
```diff
@@ -232,7 +232,8 @@ impl<IN: ArconType> Stream<IN> {
         let mut state_dir = self.ctx.app.arcon_conf().state_dir();
         let state_id = builder.state_id();
         state_dir.push(state_id.clone());
-        let backend = builder.create_backend(state_dir);
+        let backend =
```
Max-Meldrum, line 235: Right, I see the issue here.  A mistake on my part. We will need to give the Backend a String instead of 'static str when creating it. Storing the name as String inside the backends.  Not much else has to change, we will still access the String as &str.

Sanskar95: But that will result into multiple clones like before

Max-Meldrum: May result in some clones in the dataflow construction. But we should still be able to access that String as ``&str`` in Backend without doing clones?

Sanskar95: Yeah that can be done

arcon_state/src/backend/rocks/vec_ops.rs
```diff
@@ -31,6 +34,9 @@ impl VecOps for Rocks {
         let mut serialized = Vec::with_capacity(
             <usize as FixedBytes>::SIZE + protobuf::size_hint(&value).unwrap_or(0),
         );
+        #[cfg(feature = "metrics")]
+        record_bytes_written(handle.name(), serialized.len() as u64, self.name.as_str());
+
         fixed_bytes::serialize_into(&mut serialized, &1usize)?;
         protobuf::serialize_into(&mut serialized, &value)?;
```
Max-Meldrum, line 41:
```suggestion
        fixed_bytes::serialize_into(&mut serialized, &1usize)?;
        protobuf::serialize_into(&mut serialized, &value)?;
         #[cfg(feature = "metrics")]
        record_bytes_written(handle.name(), serialized.len() as u64, self.name.as_str());
```

src/dataflow/api.rs
```diff
@@ -38,8 +38,12 @@ pub struct OperatorBuilder<OP: Operator, Backend = DefaultBackend> {
 }
 
 impl<OP: Operator, Backend: arcon_state::Backend> OperatorBuilder<OP, Backend> {
-    pub(crate) fn create_backend(&self, state_dir: std::path::PathBuf) -> Arc<Backend> {
-        Arc::new(Backend::create(&state_dir).unwrap())
+    pub(crate) fn create_backend(
+        &self,
+        state_dir: std::path::PathBuf,
+        name: &'static str,
```
Max-Meldrum, line 44: Can be changed back to String as Backend now takes it as an argument.

```suggestion
        name: String
```

src/dataflow/stream.rs
```diff
@@ -171,7 +171,8 @@ impl<IN: ArconType> Stream<IN> {
...
+            builder.create_backend(state_dir, Box::leak(state_id.clone().into_boxed_str()));
```
Max-Meldrum, line 175:
```suggestion
            builder.create_backend(state_dir, state_id.clone()));
```

Max-Meldrum:
Looks good, I'll merge it!

Closes #242

PR #250, opened 2021-08-17
chore(refactor): ChannelStrategy/OperatorContext rework

### LINKED ISSUE #119: Rework Node/OperatorContext
Look into removing the RefCell's inside Node. At least remove the borrow per operator call (putting the context outside the event loop).

### LINKED ISSUE #248: Refactor ChannelStrategy/OperatorContext
The current implementation passes along a reference of ``ComponentDefinition`` across multiple functions just so that a ``ChannelStrategy`` can use it for kompacts ``tell_serialised`` function.

Refactor ``ChannelStrategy`` to return an iterator of (Channel, ArconMessage) that can be dispatched from the Node/SourceNode itself. This way we also can handle potential buffer errors easier too.

The above change also relates to operators as we also pass a ``ComponentDefinition`` reference into the ``OperatorContext``.

Rather than outputting events through ``ctx.output(event)``, make  ``handle_element/handle_timeout`` return an element iterator.
```rust
// inside Operator trait
type ElementIterator: IntoIterator<Item = ArconElement<Self::OUT>> + 'static;
```

This change is also required for it to be possible to create an operator chain (#246).

Max-Meldrum:
Closes #248 Closes #119

1. Reduce the number of generic types required for ``OperatorContext``
2. Remove a bunch of unsafe calls inside Node as Operator no longer needs a Cell
3. ChannelStrategy now returns (Channel, ArconMessage) if there is something to send...


Max-Meldrum:
The last commit changes the Operator/OperatorContext quite a bit.

1. State of an Operator now lives within a Node and can be accessed through the OperatorContext ``ctx.state()``
2. Node now holds 1 OperatorContext that is borrowed out when calling operator methods.
3. OperatorBuilder now requires a state constructor.
4. Introduce Window Index (Appender, Incremental, Arrow)
5. Redesign WindowAssigner to use WindowState struct that has assigner state + a user-defined Window Index
6. Pass OperatorBuilder struct to the NodeManager so that it can construct Operators / States during runtime.
7. Change empty ArconState from ``()`` to the ``EmptyState`` struct.
8. Arcon timer is passed to the OperatorContext as ``Box<dyn Timer<...>>`` with backend type erased.

PR #295, opened 2021-12-03
chore: bump msrv + update proto macro

Max-Meldrum:


Max-Meldrum:
@segeljakt 

tag in ``get_prost_field_attr`` seems to always be None. Should it be removed?

PR #297, opened 2021-12-14
refactoring constructor

adamhass:
This refactors the constructor which would bring the master branch closer to the distributed branch I'm working on and will also make it easier to continue working on the Keyed API.

I stupidly also took care of some warnings and such so probably touching more files than was necessary, hope that's ok.

### DIFF
--- a/arcon/arcon_state/src/backend/sled/mod.rs
+++ b/arcon/arcon_state/src/backend/sled/mod.rs
@@ -280,6 +280,7 @@ mod tests {
     use tempfile::TempDir;
 
     #[derive(Debug)]
+    #[allow(dead_code)]
     pub struct TestDb {
         sled: Arc<Sled>,
         dir: TempDir,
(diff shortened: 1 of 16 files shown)


adamhass:
Don't know what's up with those merge conflicts?

Max-Meldrum:
> Don't know what's up with those merge conflicts?

Not sure either 😄  

Looks good just have some change requests / questions.

1.  Should/Can we try to move allocator/logger to ``AssembledApplication`` too? would get rid of ``app.app.`` calls also.
2.  Commented and unused code should be removed
3.  Many eprintlns are there and could be removed.
4.  There are some changes on timeout e.g., ``wait(250)`` to ``wait(1000)``. any reason for this? take more time now?
5.  Yeah the new warnings are due to a newer stable release and Clippy. (might push a fix for this to master and it can be synced)

Max-Meldrum:
I think it would be better to sync this with our keyed work instead for now.  Then when merging that, one can worry about having it clean (whether it is printlns or commented code).

PR #299, opened 2022-01-13
Adding key_by() function to Stream<T>

adamhass:
### Adds the method key_by(...) to Stream<T>

Type signature:
```
pub fn key_by<F, KEY>(mut self, key_extractor: F) -> Stream<T>
    where
        KEY: Hash + 'static,
        F: Fn(&T) -> &KEY + ArconFnBounds,
```

### ChannelKind::Forward
Since keyed ChannelKind is actually meaningful now it is no longer the default ChannelStrategy of operators. Instead the "Forward" channel strategy is the default and it has a slightly modified behaviour. If Operator A with parallelism Pa (nodes A1, A2... APa) sends messages, using the forward strategy, to Operator B with parallelism Pb (nodes B1, B2... BPb) then node A1 sends to B1 and A2 to B2 etc. Note: The program will fail to build the graph at runtime if Pa>Pb>1. If Pb>Pa there will be unused nodes in B.

### Refactoring
Also includes refactoring of the constructor.rs file and refactoring the way Application/AssembledApplication structs.

### Testing
I suppose the key_by() feature should be more well tested eventually but I believe this is about as good of a place as any to say "good enough", as the API is done and basic functionality is tested in the new integration test file.

### DIFF
--- a/arcon/src/application/mod.rs
+++ b/arcon/src/application/mod.rs
@@ -6,31 +6,23 @@ use crate::stream::source::{
     schema::SourceSchema,
 };
 use crate::{
-    application::conf::{logger::ArconLogger, ApplicationConf, ExecutionMode},
+    application::conf::{logger::ArconLogger, ApplicationConf},
     buffer::event::PoolInfo,
-    data::ArconMessage,
     dataflow::{
         api::{ParallelSourceBuilder, SourceBuilder, SourceBuilderType},
         conf::SourceConf,
-        constructor::{source_manager_constructor, ErasedComponent},
+        constructor::SourceConstructor,
         dfg::*,
         stream::Context,
     },
-    manager::{
-        epoch::{EpochEvent, EpochManager},
-        snapshot::SnapshotManager,
-    },
     prelude::*,
-    stream::{
-        node::{debug::DebugNode, source::SourceEvent},
-        source::{local_file::LocalFileSource, Source},
-    },
+    stream::source::{local_file::LocalFileSource, Source},
 };
 use arcon_allocator::Allocator;
-use kompact::{component::AbstractComponent, prelude::KompactSystem};
+
 use std::sync::{Arc, Mutex};
 
-mod assembled;
+pub(crate) mod assembled;
 pub mod conf;
 
 pub use crate::dataflow::stream::Stream;
@@ -66,26 +58,16 @@ use metrics_exporter_prometheus::PrometheusBuilder;
 /// ```
 #[derive(Clone)]
 pub struct Application {
-    /// [`KompactSystem`] for Control Components
-    pub(crate) ctrl_system: KompactSystem,
-    /// [`KompactSystem`] for Data Processing Components
-    pub(crate) data_system: KompactSystem,
     /// Configuration for this application
     pub(crate) conf: ApplicationConf,
+    /// The DFG of this application
+    pub(crate) dfg: DFG,
     /// Arcon allocator for this application
     pub(crate) allocator: Arc<Mutex<Allocator>>,
-    /// SourceManager component for this application
-    pub(crate) source_manager: Option<Arc<dyn AbstractComponent<Message = SourceEvent>>>,
-    /// EpochManager component for this application
-    pub(crate) epoch_manager: Option<Arc<Component<EpochManager>>>,
-    /// SnapshotManager component for this application
-    pub(crate) snapshot_manager: Arc<Component<SnapshotManager>>,
+    /// A container holding information about the application's control plane
+    // pub(crate) control_plane: ControlPlaneContainer,
     /// Flag indicating whether to spawn a debug node for the Application
     debug_node_flag: bool,
-    // Type erased Arc<Component<DebugNode<A>>>
-    pub(crate) debug_node: Option<ErasedComponent>,
-    // Type erased Arc<dyn AbstractComponent<Message = ArconMessage<A>>>
-    pub(crate) abstract_debug_node: Option<ErasedComponent>,
     /// Configured Logger for the Application
     pub(crate) arcon_logger: ArconLogger,
 }
@@ -122,21 +104,24 @@ impl Application {
                 error!(arcon_logger, "metrics recorder has already been set");
             }
         }
-
-        let (ctrl_system, data_system, snapshot_manager, epoch_manager) =
-            Self::setup(&conf, &arcon_logger);
-
+        /*
+        let control_plane = match &conf.control_plane_mode {
+            ControlPlaneMode::Embedded => {
+                let mut cp_conf = ControlPlaneConf::default();
+                let mut dir = conf.base_dir.clone();
+                dir.push("control_plane");
+                cp_conf.dir = dir;
+                ControlPlaneContainer::Embedded(ControlPlane::new(cp_conf))
+            }
+            ControlPlaneMode::Remote(addr) => ControlPlaneContainer::Remote(addr.clone()),
+        };
+        */
         Self {
-            ctrl_system,
-            data_system,
             conf,
+            dfg: DFG::default(),
             allocator,
-            snapshot_manager,
-            epoch_manager,
-            source_manager: None,
+            //control_plane,
             debug_node_flag: false,
-            debug_node: None,
-            abstract_debug_node: None,
             arcon_logger,
         }
     }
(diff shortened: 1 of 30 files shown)


Max-Meldrum:
Nice work! 😄  I have a few change requests for this PR, but then other things that we can create separate issues for.

arcon/src/application/mod.rs
```diff
@@ -66,26 +58,16 @@ use metrics_exporter_prometheus::PrometheusBuilder;
...
     /// Arcon allocator for this application
     pub(crate) allocator: Arc<Mutex<Allocator>>,
-    /// SourceManager component for this application
-    pub(crate) source_manager: Option<Arc<dyn AbstractComponent<Message = SourceEvent>>>,
-    /// EpochManager component for this application
-    pub(crate) epoch_manager: Option<Arc<Component<EpochManager>>>,
-    /// SnapshotManager component for this application
-    pub(crate) snapshot_manager: Arc<Component<SnapshotManager>>,
+    /// A container holding information about the application's control plane
+    // pub(crate) control_plane: ControlPlaneContainer,
```
Max-Meldrum, line 68:
```suggestion

```

arcon/src/application/mod.rs
```diff
@@ -122,21 +104,24 @@ impl Application {
...
+            ControlPlaneMode::Embedded => {
+                let mut cp_conf = ControlPlaneConf::default();
+                let mut dir = conf.base_dir.clone();
+                dir.push("control_plane");
+                cp_conf.dir = dir;
+                ControlPlaneContainer::Embedded(ControlPlane::new(cp_conf))
+            }
+            ControlPlaneMode::Remote(addr) => ControlPlaneContainer::Remote(addr.clone()),
+        };
+        */
```
Max-Meldrum, line 118:
```suggestion
```

arcon/src/application/mod.rs
```diff
@@ -122,21 +104,24 @@ impl Application {
...
         Self {
-            ctrl_system,
-            data_system,
             conf,
+            dfg: DFG::default(),
             allocator,
-            snapshot_manager,
-            epoch_manager,
-            source_manager: None,
+            //control_plane,
```
Max-Meldrum, line 123:
```suggestion
```

arcon/src/dataflow/constructor.rs
```diff
@@ -30,306 +32,482 @@ use arcon_state::Backend;
...
-        .start_notify(&source_node_comp)
-        .wait_timeout(std::time::Duration::from_millis(2000))
-        .expect("Failed to start Source Node");
-
     biconnect_components::<SourceManagerPort, _, _>(source_manager_comp, &source_node_comp)
         .expect("failed to biconnect components");
 
+    app.data_system().start(&source_node_comp);
+    //.wait_timeout(std::time::Duration::from_millis(5000))
+    //.expect("Failed to start Source Node");
```
Max-Meldrum, line 506: Should this be commented out?

arcon/src/dataflow/dfg.rs
```diff
@@ -13,124 +37,116 @@ pub struct DFG {
...
+        self.graph.get_mut(*id).unwrap()
     }
 }
 
-/// The ID of a [`DFGNode`] in the dataflow graph.
-#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
-pub struct DFGNodeID(pub usize);
+// /// The ID of a [`DFGNode`] in the dataflow graph.
+// #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+// pub struct OperatorId(pub usize);
```
Max-Meldrum, line 61:
```suggestion

```

arcon/src/dataflow/stream.rs
```diff
@@ -1,52 +1,203 @@
...
+    /// must return the same key whenever it is called.
+    ///
+    /// Example
+    /// ```no_run
+    /// use arcon::prelude::*;
+    /// let stream: Stream<u64> = Application::default()
+    ///     .iterator(0u64..100, |conf| {
+    ///         conf.set_arcon_time(ArconTime::Process);
+    ///     })
+    ///     .key_by(|i: &u64| i);
```
Max-Meldrum, line 87:
```suggestion
    ///     .key_by(|i: &u64| i);
    /// ```
```

arcon/src/dataflow/stream.rs
```diff
@@ -1,52 +1,203 @@
...
+    /// Sets the [ChannelKind] of the last operator. The default [ChannelKind] is [ChannelKind::Forward].
+    ///
+    /// Example
+    /// ```no_run
+    /// use arcon::prelude::*;
+    /// let stream: Stream<u64> = Application::default()
+    ///     .iterator(0u64..100, |conf| {
+    ///         conf.set_arcon_time(ArconTime::Process);
+    ///     })
+    ///     .channel_kind(ChannelKind::Forward);
```
Max-Meldrum, line 123:
```suggestion
    ///     .channel_kind(ChannelKind::Forward);
    /// ```
```

arcon/src/test/keyby_integration.rs
```diff
@@ -0,0 +1,160 @@
...
+            }),
+            conf: operator_conf(),
+        })
+}
+
+#[test]
+fn key_by_integration() {
+    let mut app = enriched_event_stream().build();
+    app.start();
+    sleep(Duration::from_secs(10));
```
Max-Meldrum, line 103: Sleeping 10 seconds seems like a lot.  Increases total testing time

arcon/src/test/keyby_integration.rs
```diff
@@ -0,0 +1,160 @@
...
+                Map::new(|event: EnrichedEvent| event)
+            }),
+            state: Arc::new(|_| EmptyState),
+            conf: operator_conf(),
+        })
+        .build();
+
+    app.start();
+
+    sleep(Duration::from_secs(10));
```
Max-Meldrum, line 141: Same here. 10 seconds is a lot.

arcon/src/stream/channel/strategy/keyed.rs
```diff
@@ -92,11 +90,12 @@ where
...
         match &event {
-            ArconEvent::Element(element) => {
+            ArconEvent::Element(e) => {
                 // Get key placement
-                let key = element.data.get_key() % self.max_key;
+                //let key = self.extractor(&element.data);
+                let key = self.key_builder.get_key(&e.data);
                 // Calculate which key range index is responsible for this key
-                let index = (key * self.key_ranges / self.max_key) as usize;
+                let index = (key % self.key_ranges) as usize; // ??
```
Max-Meldrum, line 98: With the updated partitioning code in https://github.com/cda-group/arcon/blob/master/arcon/src/data/partition.rs , this is where we should do the shard lookup after extracting the key. 

As we already have extracted the key, I reckon we might have to add another function or refactor ``shard_lookup``.

```rust
let key = self.key_builder.get_key(&e.data);
let shard/index = shard_lookup_with_key(key, self.total_shards);
```
Perhaps something like this.

```rust
// src/data/partition.rs
pub fn shard_lookup<K>(data: &K, total_shards: u32) -> u32
where
    K: Hash + ?Sized,
{
    let mut hasher = arcon_util::key_hasher();
    data.hash(&mut hasher);

    shard_lookup_with_key(hasher.finish())
}

#[inline]
pub fn shard_lookup_with_key(hashed_key: u64, total_shards: u32) -> u32 {
  let key = hashed_key % MAX_KEY;
  key * total_shards / MAX_KEY
}
```

arcon/src/stream/operator/function/mod.rs
```diff
@@ -44,7 +44,7 @@ mod tests {
     // helper to check common result between Map/MapInPlace
     fn check_map_result(mut app: AssembledApplication) {
         app.start();
-        wait(250);
+        wait(1000);
```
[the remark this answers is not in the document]
adamhass: Yeah flakiness

arcon/src/stream/channel/strategy/keyed.rs
adamhass: Ok I used that snippet. This Shard/Key/Index/MAX_KEY/total_shards needs documenting but I can't do it, it's very confusing to me.

Max-Meldrum: Yes, that is a fair point. The new shard terminology and its functions have not yet been integrated with the runtime. 
So there is a bit of mix of terms used (e.g., shard vs. index)

And yes, it does indeed need more explaination text. Can leave that for other issue.

In short.

MAX_KEY represents the max key 0...65363

first calculation is to find where the hashed key maps to in the key space (0..MAX_KEY). Then the next step
is to identify the shard responsible for that key.

But yeah let's leave the integration and documentation for other issue/pr. But could you remove the PartitionIndex type? Leaving it as u32/u64 instead.

